PR #4, opened 2015-12-12
Flysystem v0.4.5 (newer versions require php >= 5.4)

dgt41:
#### Flysystem has already been named as the replacement of JFilesystem

This implementation is for local use with the new media manager (due to the use of the adapter)
The actual implementation should also take care of all the Jfile and Jfolder function


roland-d:
@dgt41 Is this still relevant since we won't be incorperating Flysystem at the moment?

yireo:
No, this PR can be closed.

PR #6, opened 2015-12-28
WIP connect the tree view with the files view with ajax pt1

dgt41:


dgt41:
@jissereitsma we will need to change some parts of the current models to get more data available for the folders and the files. Eg for the folders will be way easier if we have the depth (relative to /images or the root images directory) the path for the directory above (for the back button) and for the files the type and the layout path (these should also be injected from the plugin that will extend the file types available). But since you did all the work on the models I guess it will be better if we can Skype and work out these ideas, so we can move this forward

jissereitsma:
@dgt41 Were you able to attend this afternoon on Skype? We had a little meeting planned on the progress, though some people can't make it. Let's discuss this then.

The folder model would need an extra `depth` attribute.

The files are already including the right information, at least with my current PR (still pending merge, just like this one): Each file has a `file_type` (referring to `image`, or `pdf` or something else) which can be extended using plugins. And the file type simply determines the sublayout (`views/files/tmpl/default_pdf.php`) which is also included in my own PR.

roland-d:
@dgt41 I can't merge this due to conflicts. Please check and let me know.

PR #8, opened 2016-04-10
Another merge to have working version

yireo:
This commit fixes many issues, and will demo a working version of the new MM so people can join and improve things from here.


jissereitsma:
It's just me.

PR #12, opened 2016-05-28
Refactoring: improves if structures (less indentation)

regularlabs:
Also removes a redundant if check (line 74) as it is already checked earlier.


jissereitsma:
Good call. Note that the whole project is still under a lot of changes. Code compliance might come in later. And you're welcome to join us in a code sprint end of June somewhere in Weert ;)

PR #16, opened 2016-05-28
Refactor: decreases indentations via early returns and extracting code into separate methods

regularlabs:


jissereitsma:
Just nice

PR #18, opened 2016-08-01
Imagecropper fixes

roland-d:
#### Summary of Changes

Updated the imagecropper library to the latest version
Fixed the toolbar buttons
Cleaned up the file to use Joomla conventions
Fixed Javascript functions
#### Testing Instructions

There are two ways to test any PR here:
- The easy one where you need to download a copy of this repo from [here](https://github.com/joomla-projects/media-manager-improvement/archive/master.zip) and then install it locally. Also grab com_patchtester and install that too and then go to it's configuration options and set the project name to joomla-projects and the repo to media-manager-improvement
  After that a fetch in the com_patchtester will show all the PRs and you can apply any of those and test it.
- The hard way needs some git commands where you get that repo locally and then switch to any branch. Again you will have to install it locally pointing your servers path to the folder that the local repo exists 

Basically, try to crop an image and see if it works.
#### NOTE:

Right now there is an inconsistency in the upload logic (php side) for files with a `-` which will be converted to `_`. So please check the created files carefully
#### Final remarks

Although this feature isn't fully crystallized I think it is usable as-is for testing.


jissereitsma:
Looks solid to me: Some language strings as well. However with that, I'm not sure what the standards are: You have renamed the language prefix from `PLG_MEDIA-EDITOR_IMAGECROPPER` to `PLG_IMAGECROPPER` so removing the plugin group. Isn't that wrong? In the core, the plugin group is used everywhere. To me it sounds like a good idea as well to include the group.

marcodings:
I don't think its wrong in the sense its not allowed. Iirc the the standards only call for "name spacing".
Having said that i second the notion to keep the PLG_MEDIA-EDITOR_ grouping for readability reasons

plugins/media-editor/imagecropper/layout/form.php
```diff
@@ -6,32 +6,37 @@
...
     <?php echo JHtml::_('image', COM_MEDIA_BASEURL . '/' . $filePath, '', 'id="media-image"'); ?>
 </div>
 <br />
-<input type="checkbox" name="save_copy" id="save_copy" checked="checked"/> <label for="save_copy"><?php JText::printf('PLG_MEDIA-EDITOR_IMAGECROPPER_SAVE_COPY') ?></label>
-<input type="submit" name="<?php echo JText::_('PLG_MEDIA-EDITOR_IMAGECROPPER_FIELD_SUBMIT'); ?>" />
+<input type="checkbox" name="save_copy" id="save_copy" checked="checked"/> <label for="save_copy"><?php echo JText::_('PLG_IMAGECROPPER_SAVE_COPY') ?></label>
+<input type="submit" class="btn" value="<?php echo JText::_('PLG_IMAGECROPPER_FIELD_SUBMIT'); ?>" />
 <input type="hidden" name="imagecropper-jsondata" value="" id="imagecropper-jsondata" />
 
 <script type="text/javascript">
```
dgt41: @roland-d can you create a `media/plg_media-editor_imagecropper/js/cropper-init.js` and paste the next lines in it? Having inline script tags -especially in the body tag- is not good.
probably wrapped in 
```js
!function() {
// code here
}(); // self executing function
```

wilsonge:
I don't think there should be hypens in our language strings. I'd replace the hypen with an underscore. But otherwise i agree we should keep the plugin group

brianteeman:
I'm with @wilsonge in not having any hyphens in a language string

brianteeman:
I notice in the strings you are calling this Imagecropper which isnt a word in en-GB
I would go with the same as the original library and call it "Image Cropper"

roland-d:
I was confused with what to do with the language strings because most don't have a hyphen but the editors-xtd group does have a hyphen in its name. So what to do? Joomla itself requires a hyphen for the plugin name because the folder has a hyphen so we can't remove them completely.

As for the group, the editors-xtd language files are pretty inconsistent as well. I am fine with adding the group and use an underscore. The inconsistencies in the editors-xtd can't be fixed now anyway I guess.

Thanks Brian, I have updated the name as suggested.

brianteeman:
Valid point about the editors-xtd strings and I dont like those either but its too much of a fight to change those now.  :(

It would be much easier to read if we just use underscores in the strings (I dont care about the filenames)

roland-d:
At least we can do it right for these ones. As I said some hyphens have to stay due to the plugin folder name having hyphens. @jissereitsma What do you say we change the folder name as well to have underscores? Just so we can be consistent.

brianteeman:
Peter was talking about trying to document some standards for strings when we were in manchester https://github.com/joomla/user-interface-text/issues/61

plugins/media-editor/imagecropper/layout/form.php
roland-d: @dgt41 I tried but it breaks the Javascript, so will need your help with this.

media/plg_media-editor_imagecropper/js/cropper-init.js
```diff
@@ -0,0 +1,61 @@
...
+          "scaleX": e.detail.scaleX,
+          "scaleY": e.detail.scaleY
+        }];
+
+        document.getElementById('imagecropper-jsondata').value = json;
+      }
+    });
+    // We are using jQuery only to add a click event to the buttons and the dom ready? We are doing it wrong!!
+    jQuery('.btn-toolbar button').click(function () {
+      var action = jQuery(this).data('method');
```
dgt41: The native javascript here is:
```js
jQuery('.btn-toolbar button').click(function (event) {
     var action = event.currentTarget.getAttribute("data-method");
```

roland-d:
@dgt41 The save as copy option is gone. Why did you remove it?

plugins/media-editor/imagecropper/layout/form.php
```diff
@@ -1,62 +1,79 @@
...
 <div class="cropper-bg">
-    <?php echo JHtml::_('image', COM_MEDIA_BASEURL . '/' . $filePath, '', 'id="media-image"'); ?>
+    <?php echo JHtml::_('image', COM_MEDIA_BASEURL . '/' . $filePath, '', 'id="joomla-media-image-cropper" data-some-attribute="some value" data-url="'. $uploadUrl . '"'); ?>
 </div>
 <br />
-<input type="checkbox" name="save_copy" id="save_copy" checked="checked"/> <label for="save_copy"><?php JText::printf('PLG_MEDIA-EDITOR_IMAGECROPPER_SAVE_COPY') ?></label>
-<input type="submit" name="<?php echo JText::_('PLG_MEDIA-EDITOR_IMAGECROPPER_FIELD_SUBMIT'); ?>" />
+<!-- @TODO: Implement this with javascript, eg parse the old filename and then add a number at the end
+            or something similar, maybe ajax call to see if that filename already exists
+<input type="checkbox" name="save_copy" id="save_copy" checked="checked"/> <label for="save_copy"><?php echo JText::_('PLG_MEDIA_EDITOR_IMAGECROPPER_SAVE_COPY') ?></label>
```
dgt41: @roland-d read the comment here, this needs to be implemented with js, we are not doing anything in PHP with this plugin (apart the fact that we call the controller to re-upload the file).

roland-d: I am probably missing the point but why must this be Javascript? What is wrong using PHP for this? We must call the image model to make sure the image is stored in the database as well, not just saved on disk.

administrator/components/com_media/controllers/file.json.php
```diff
@@ -56,6 +56,7 @@ public function upload()
 		// Get some data from the request
 		$file   = $this->input->files->get('files', '', 'array');
 		$folder = $this->input->get('folder', '', 'path');
+		$canOverwrite = $this->input->getBool('allowovewrite', false);
```
dgt41: @roland-d The other option for creating a copy is to have a flag (e.g. createNew) and deal with it in the controller logic (this one I think is a better/easier solution)

plugins/media-editor/imagecropper/layout/form.php
dgt41: then it needs some extra logic in the controller, read my comment below (com_media/controllers/file.json.php)

administrator/components/com_media/controllers/file.json.php
roland-d: That is indeed what I am talking about, that is what the old code did. The user could choose if the cropped image would overwrite the existing one or be saved as a copy.

P.s. there is a typo in the option name. It says allowovewrite instead of allowoverwrite. Notice the r ;)

dgt41: Also this file needs some major refactoring, as I found some inconsistent behaviour in the method that checks the filename

roland-d: This whole plugin needs a major cleanup anyway and I started part of it but not sure who is going to do what now. Are you taking over?

jissereitsma:
Work in progress so merging.

PR #55, opened 2016-11-30
Add the media file

laoneo:
The media file helper class


zero-24:
Can we add a file doc block at the top?

jissereitsma:
We can have docblocks later. There is still too much moving with this project to say we need to polish things. Things can be polished when the final code is in sight, I would say.

zero-24:
Ok just ping me if you at that state. I hope i can find time than to help on that :)

jissereitsma:
Thanks. There will be so much code moved and removed, that adding those blocks now would be a waste of time. But many thanks for the offer! We'll be in touch!

PR #57, opened 2016-11-30
Adds categories and file entites code

laoneo:
This PR adds the category abstraction layer and file entities. At the moment it doesn't add any functionality to the existing media manager.


jissereitsma:
Cool. Note that we already had some file entities in place (both SQL file and file model + table), so I'm not sure how this fits in? As for the controller structure, I really want to get rid of the `controller.php` file with its weird `$vName` structure, and move it into subcontrollers. But that's for later. Work in progress ;) Awesome commit.

PR #60, opened 2016-12-01
Started integration of XHR processing controller

yvesh:


administrator/components/com_media/controllers/requests.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * The Request Controller
+ * @todo  Webservice
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class MediaControllerRequests extends JControllerLegacy
+{
+	/**
+	 * Called by the plugin through java
```
marcodings: java ;)

PR #68, opened 2016-12-23
Show the new media manager for browsers that do support ECMAScript 5

dneukirchen:
Pull Request for Issue # .


administrator/components/com_media/media.php
```diff
@@ -15,32 +17,44 @@
...
-if (!$user->authorise('core.manage', 'com_media') && (!$asset or (!$user->authorise('core.edit', $asset)
-	&& !$user->authorise('core.create', $asset)
-	&& count($user->getAuthorisedCategories($asset, 'core.create')) == 0)
-	&& !($user->id == $author && $user->authorise('core.edit.own', $asset))))
+if (!$user->authorise('core.manage', 'com_media') && (!$asset || (!$user->authorise('core.edit', $asset)
+			&& !$user->authorise('core.create', $asset)
+			&& count($user->getAuthorisedCategories($asset, 'core.create')) == 0)
+		&& !($user->id == $author && $user->authorise('core.edit.own', $asset)))
+)
 {
```
laoneo, line 24: No need to more intend the code

administrator/components/com_media/media.php
```diff
@@ -15,32 +17,44 @@
...
 
-$params = JComponentHelper::getParams('com_media');
+// Show the new media manager for browsers that do support ECMAScript 5
+$client   = JFactory::getApplication()->client;
+$basePath = JPATH_COMPONENT_ADMINISTRATOR;
 
-// Load the helper class
-JLoader::register('MediaHelper', JPATH_ADMINISTRATOR . '/components/com_media/helpers/media.php');
+if ($client->browser == WebClient::IE && $client->version <= 8)
+{
```
laoneo, line 33: For clarification, I would add here a comment, why we do something special for IE 8 and below

dneukirchen: done

administrator/components/com_media/media.php
dneukirchen: done

PR #70, opened 2016-12-27
Basic vue-app-setup

dneukirchen:
This PR includes my suggestion for a vue2 based application in the joomla component environment.

It also includes a basic list view using vue2 and the github file api (since we dont have an api yet).

### Test instructions

* Just open the media manager component in the backend

### Developmnt instructions

* Change directory to ```administrator/components/com_media```
* Run ```npm install``` to install the dependencies (only once on initial setup)
* Make your changes to the frontend application 
* Run ```npm run build```to build the app (this will generate a single javascript file ```media/media/js/app.js```)


laoneo:
What for an environment do we need when we want to do then development on that app? How do we set up this environment? If possible update the description of the PR with that information that reviewers can set up their dev machine to test this PR. Thanks.

laoneo (commented):
What for an environment do we need when we want to do then development on that app? How do we set up this environment? If possible update the description of the PR with that information that reviewers can set up their dev machine to test this PR. Thanks.

dneukirchen:
Updated description

administrator/components/com_media/resources/components/app.vue
```diff
@@ -0,0 +1,76 @@
...
+            getContent() {
+                let url = this.baseUrl + this.dir;
+                jQuery.getJSON(url, (content) => {
+                    // Update the current directory content
+                    this.content = content;
+                    // Find the directory node by path and update its children
+                    this._updateLeafByPath(this.tree, this.dir, content);
+                }).error(() => {
+                    alert("Error loading directory content.");
+                })
```
laoneo, line 39: It would be interesting to see how we will bring in multilingual support into the vue app.

dneukirchen: This is not the final solution for notifications & error handling. Its just a placeholder.
We have to implement vue notifications in a more abstract way, so that we can use it across components, services and views.

For multilang support we will use something like https://github.com/kazupon/vue-i18n and use the default joomla language files.

yvesh: @dneukirchen 

We can also use Joomla core js library here.

Joomla.JText._('COM_MEDIA_LANGUAGE_STRING');

and add them with JText::script.. (I would load a whole language file into it, so we don't need to care on that)

dneukirchen: Yes... that is also a possible solution.

I created an issue for the multilang support: issue #73

dgrammatiko:
All .vue and js files are STATIC files and need to be moved to /media/media/js and /media/media/vue respectively. We are doing it wrong here 😕

EDIT: Ignore this comment, there is some grunt magic that I overlooked

PR #71, opened 2016-12-28
Api controller endpoints

laoneo:
Api controller to work with the local file system. The returned data is always json created by a Joomla json response with the format:

    {
        "success":true,
        "message":"ok",
        "messages":null,
        "data":[
            {
                "type":"dir",
                "name":"banners",
                "path":"//"
            },
            {
                "type":"dir",
                "name":"headers",
                "path":"//"
            },
            {
                "type":"file",
                "name":"joomla_black.png",
                "path":"//"
            },
            {
                "type":"file",
                "name":"powered_by.png",
                "path":"//"
            }
        ]
    }

The following HTTP methods are handled:
- GET returns folder and file information
- POST is not implemented yet
- PUT is not implemented yet
- DELETE deletes a folder or file

The controller delegates the actions to the configured adapter. Some basic unit tests are done for the local adapter.


administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -0,0 +1,132 @@
...
+			$path = '/' . trim($path, '/') . '/';
+		}
+
+		// The data to return
+		$data = array();
+
+		// Read the folders
+		foreach (JFolder::folders($this->rootPath . $path) as $folder)
+		{
+			$obj       = new stdClass();
```
yvesh: Remove Parenthesis () - just new stdClass;

administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -0,0 +1,132 @@
...
+			$data[]    = $obj;
+		}
+
+		// Read the files
+		foreach (JFolder::files($this->rootPath . $path) as $file)
+		{
+			$obj       = new stdClass();
+			$obj->type = 'file';
+			$obj->name = $file;
+			$obj->path = $path;
```
yvesh: How about extending that with:

```php
$obj->extension = pathinfo($path, PATHINFO_EXTENSION);
$obj->length    = filesize(..);
```

administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -0,0 +1,132 @@
...
+			$success = JFile::delete($this->rootPath . $path);
+		}
+		else
+		{
+			$success = JFolder::delete($this->rootPath . $path);
+		}
+
+		if (!$success)
+		{
+			throw new Exception('Delete not possible!');
```
yvesh, line 145: Should be catched above then.

laoneo: Where above, in the controller?

administrator/components/com_media/libraries/media/file/adapter/local.php
laoneo: Done

administrator/components/com_media/libraries/media/file/adapter/local.php
yvesh: Yep and than add a user friendly message

Btw. any way to delete multiple files at once now?

dneukirchen: I totally agree with @yvesh, but i would suggest sending a json response with the error flag set to true, a user firendly message (and optionally the correct response status code). 

I dont like JResponseJson, but i think we have to use it...

EDIT: i saw that you´re already catching the adapter exceptions... So its fine for me,

dneukirchen (approved):
Thanks Allon. Good job.

administrator/components/com_media/controllers/api.php
```diff
@@ -0,0 +1,150 @@
...
+ * This is NO public api controller, it is internal for the com_media component only!
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class MediaControllerApi extends JControllerLegacy
+{
+	/**
+	 * The local file adapter to work with.
+	 *
+	 * @var MediaFileAdapterInterface
```
yvesh: cs and since :)

administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -0,0 +1,143 @@
...
+ * Local file adapter.
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class MediaFileAdapterLocal implements MediaFileAdapterInterface
+{
+	/**
+	 * The root path to gather file information from.
+	 *
+	 * @var string
```
yvesh, line 25: cs and since

administrator/components/com_media/controllers/api.php
```diff
@@ -0,0 +1,150 @@
...
+	 *
+	 * @param   mixed    $data     The data to send
+	 * @param   string   $message  The message
+	 * @param   boolean  $error    If it is an error response
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function sendAndClose($data, $message, $error)
```
yvesh: General question, should we stay on protected for all intern methods? So we still have the ability to override them?

yvesh (commented):
Thanks Allon! :)

administrator/components/com_media/controllers/api.php
dneukirchen: +1 for protected methods **AND** properties

administrator/components/com_media/libraries/media/file/adapter/local.php
laoneo: Yes, it's library, so throwing exceptions should be fine here. The controller is responsible to handle the exception and to create a json response.

laoneo:
I'v used in the past with DPCalendar also a format specific controller like api.json.php. The problem is when a Joomla site still has the .htaccess file from 2.5 in use, then this will fail on the front end as they have have the formats hardcoded. I had to revert it back to format raw.

I think we should also be prepared to have it working on the front end.

administrator/components/com_media/controllers/api.json.php
```diff
@@ -0,0 +1,145 @@
...
+	 * Api endpoint for the media manager front end. The HTTP methods GET, PUT, POST and DELETE
+	 * are supported.
+	 *
+	 * The following query parameters are processed:
+	 * - path: The path of the resource, if not set then the default / is taken.
+	 *
+	 * Some examples with a more understandable rest url equivalent:
+	 * - GET a list of folders below the root:
+	 * 		index.php?option=com_media&task=api.files
+	 * 		/api/files
```
dgrammatiko, line 64: Am I the only one that thinks this is totally wrong?
- The route /api should be followed from a version e.g. /v1/
- Using /api for media manager will conflict very badly with the web services, if ever get to a stable version
- I suggest a /media/... (index.php?option=com_media&task=media.files) so a media.json.php filename and MediaControllerMedia class

dgrammatiko: Also another essential part of a good API is throttling, so I guess we can implement some simple version with just an IP check and a simple call to db (maybe using the session table)

PR #76, opened 2017-01-07
Add the process and preview actions

laoneo:
Adds support to process a media file trough the api controller. A preview can also be requested for post parameters.

This pr is based on #75.

PR #77, opened 2017-01-09
Simplify the process event, sending the local file path

laoneo:
In the original pr #76 the _onMediaProcess_ event is triggered with a resource object. This is very much tailored to images. This pr uses a local file system path instead.

Pros:
- It will make the event more understandable for developers.
- Libraries which are file based are supported.

Cons:
- Every plugin does probably a file open and close.


laoneo:
Closing as we have a different strategy https://volunteers.joomla.org/teams/new-media-manager-team/reports/427-media-manager-status-meeting-19-1-2017

Issue #89, opened 2017-02-22
API / Fileadapter: resource response data for create folder request


dneukirchen:
### Steps to reproduce the issue

Try to create a new folder:
POST /administrator/index.php?option=com_media&format=json&task=api.files&path=/
JSON BODY: {name: 'my-folder'}

### Expected result
A single resource response (not an array of items) with the data of the newly created folder:
```
{
  success: true,
  message: null, 
  messages: null, 
  data: {
    type: "folder", 
    name: "my-folder", 
    path: "/my-folder", 
    extension: "", 
    size:…
  }
}
```

### Actual result
The contents of the newly created folder. 
``` 
{success: true, message: null, messages: null, data: []}
```

dneukirchen:
Quick and dirty solution: https://github.com/joomla-projects/media-manager-improvement/commit/7f87259eab162de41202d20d46ebc5877f1ce752

laoneo:
This should be fixed with PR #96.

Issue #73, opened 2016-12-28
Implement multilanguage support in the vue frontend application


dneukirchen:
* Use the default joomla language files 

Vue Plugins:

* https://github.com/kazupon/vue-i18n

laoneo:
We can also use the Joomla way `Joomla.JText._('COM_MEDIA_LANGUAGE_STRING');`.

dgrammatiko:
I will also support @laoneo way, for the default joomla language files

yvesh:
I told you @dneukirchen :)

dneukirchen:
With plugins like https://github.com/kazupon/vue-i18n we will also use the default joomla language files (.ini´s). 

I dont like Joomla.JText and i think the technique is not ment to be used with powerful tools like vue, angular and react. It feels more like the quick and dirty solution, but i will give in to the majority.

mbabker:
The "issue" with `Joomla.JText` is it requires you to preload strings on the PHP side so they can be used with a request, it's not designed for a JavaScript frontend at all like a Vue or Angular or React.  Honestly, it'd be nice if we could do something similar to https://github.com/willdurand/BazingaJsTranslationBundle that creates a way to expose a language catalogue to the JS API, but that would also mean making `JLanguage` smart enough to be able to create a compiled catalogue with all the strings (and that's also impossible because of extensions overriding strings defined in other files so some of the `JLIB_*` strings have different text based on the active extension versus having a proper one key per string globally definition; funny enough I've got code to build the cached catalogue but it gets refreshed on so many page loads because of key overloading it's useless).

Though to be honest, Joomla as it is built isn't really designed to be used with a JavaScript application framework.  So none of the existing tooling will just "plug" into a totally new application architecture.  There's no winning solution to that issue.

laoneo:
But are the ini files not automatically loaded on the first page request? And the ajax requests should return only translated data anyway. So why should it be a problem to use JText then from technical point of view?

I'm aware that there are better ways to bring in translations, but JText is the Joomla way and there will be a learning curve when we go with the vue approach, so when we can use existing Joomla functionality, then I think it twill make the entrance for others easier.

dneukirchen:
I agree with @mbabker, but i think the feature (generate a compiled language string catalogue) is out of the scope of this project and could be a project on its own (especially when you have to find a clean solution for the overloading issue).

And for JText vs. vue i18n plugin: 
[...]

PR #90, opened 2017-02-22
Feature/create folder modal

dneukirchen:
### Summary of Changes

With this PR it is possible to create new folders in the currently selected directory.

bildschirmfoto 2017-02-22 um 22 47 
25

* Created a reusable modal component with the look and feel of b2 modals
* Created create-folder-modal component (which uses the modal component)
* Created the "create-folder-form" inside the modal
* Updated the client api: implemented createFolder method
* Updated state actions and mutations with the new possibilities
* Quick Fix for the missing single resource response in the backend api and filehandler

Follow up TODOs:
* Modal styling
* Update backend api (it was just a quick fix for the resource response see #89)
* Foldernames (currently all whitespaces are stripped, perhaps it is better to replace them with '-')
* Put the create-folder-form state in vuex
* Transition effects for the modal

### Testing Instructions

* Click on create folder
* Create folder


administrator/components/com_media/controllers/api.json.php
```diff
@@ -144,7 +144,7 @@ public function files()
 						$this->adapter->createFolder($name, $path);
 					}
 
-					$data = $this->adapter->getFiles($path . '/' . $name);
+					$data = $this->adapter->getFile($path . '/' . $name);
```
dgt41: You can have multiple files uploaded...

dneukirchen: But not in the current implementation if the api. The adapter calls create only one file/folder per request.

```
$this->adapter->createFile($name, $path, $mediaContent);

$this->adapter->createFolder($name, $path);
```

administrator/components/com_media/resources/components/toolbar/toolbar.vue
```diff
@@ -2,7 +2,7 @@
     <div class="media-toolbar">
         <div class="create-wrapper">
             <div class="btn-group">
-                <button class="btn btn-success">Upload</button>
+                <button class="btn btn-success">New</button>
```
zero-24, line 5: No translation planed?

laoneo: https://github.com/joomla-projects/media-manager-improvement/issues/73

dgrammatiko: https://github.com/dgt41/bs4-custom-elements/blob/master/elements/alert/alert.js#L64-L76

Issue #82, opened 2017-02-17
[API] Handle invalid path requests


dneukirchen:
### Steps to reproduce the issue
Make a get request with an invalid path 

```GET /administrator/index.php?option=com_media&format=json&task=api.files&path=some-invalid---or-not-existing-path```

### Expected result
Server responds with an 404 error.

### Actual result
Server responds with an 200 error and empty data set

laoneo:
Closing as we have a pr #96.

PR #96, opened 2017-02-23
Return error codes when an exception happens

laoneo:
Pull Request for Issue #82.

Sends a response code when an exception happens. 404 when the path is invalid and 500 when a general exception happens.


dneukirchen:
lgtm.

In an optimal world status code and content-type header would be set by JResponseJson, but the class is not capable of doing it, isnt it?

dneukirchen (approved):

laoneo:
No it's not.

PR #100, opened 2017-02-24
Use JText to translate strings

laoneo:
Pull Request for Issue #73.

### Summary of Changes
A new plugin which adds a global mixin to translate strings. The strings are loaded from JText.

In a component you can use `{{ translate('COM_MEDIA_NEW') }}` which will then translate the string from the joomla ini file.


dneukirchen (approved):
lgtm. Thank you allon.

PR #101, opened 2017-02-26
Feature/joomla acceptance

dneukirchen:
### Summary of Changes

Lets face it: we have enough problems getting acceptance and approval for our tec-stack, so lets not go wild with the ui/layout stuff and do that part as joomla´ish as possible. 😢 😞 

Even though i dont like the look and feel of the current j4 backend theme, this PR tries to make the nmm look like a default joomla 4 component.

* Removed our custom styles. Using b4/j4 classes now.
* Re-added the ugly joomla subhead toolbar and connected the buttons with Vue using Eventbus pattern.
* Removed breadcrumb from toolbar and added it above the media-browser
* Style adjustements.

Preview: 
bildschirmfoto 2017-02-26 um 13 10 05

Before:
bildschirmfoto 2017-02-26 um 13 25 05

### DIFF
--- a/administrator/components/com_media/layouts/toolbar/create-folder.php
+++ b/administrator/components/com_media/layouts/toolbar/create-folder.php
@@ -9,6 +9,6 @@
 defined('_JEXEC') or die;
 $title = JText::_('COM_MEDIA_CREATE_NEW_FOLDER');
 ?>
-<button class="btn btn-sm btn-outline-info">
-	<span class="icon- fa fa-folder" title="<?php echo $title; ?>"></span> <?php echo $title; ?>
+<button class="btn btn-sm btn-outline-info" onclick="MediaManager.Event.$emit('onClickCreateFolder');">
+    <span class="icon- fa fa-folder" title="<?php echo $title; ?>"></span> <?php echo $title; ?>
 </button>
--- a/administrator/components/com_media/layouts/toolbar/create-folder.php
+++ b/administrator/components/com_media/layouts/toolbar/create-folder.php
@@ -1,0 +1,14 @@
+<?php
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_media
+ *
+ * @copyright   Copyright (C) 2005 - 2017 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+defined('_JEXEC') or die;
+$title = JText::_('COM_MEDIA_CREATE_NEW_FOLDER');
+?>
+<button class="btn btn-sm btn-outline-info">
+	<span class="icon- fa fa-folder" title="<?php echo $title; ?>"></span> <?php echo $title; ?>
+</button>
--- a/administrator/components/com_media/layouts/toolbar/delete.php
+++ b/administrator/components/com_media/layouts/toolbar/delete.php
@@ -1,0 +1,14 @@
+<?php
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_media
+ *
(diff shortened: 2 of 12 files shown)


administrator/components/com_media/layouts/toolbar/create-folder.php
```diff
@@ -0,0 +1,14 @@
...
+ * @subpackage  com_media
+ *
+ * @copyright   Copyright (C) 2005 - 2017 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+defined('_JEXEC') or die;
+$title = JText::_('COM_MEDIA_CREATE_NEW_FOLDER');
+?>
+<button class="btn btn-sm btn-outline-info" onclick="MediaManager.Event.$emit('onClickCreateFolder');">
+    <span class="icon- fa fa-folder" title="<?php echo $title; ?>"></span> <?php echo $title; ?>
```
yvesh, line 13: remove the icon- and use why span and not i (like common in bs?)

administrator/components/com_media/layouts/toolbar/delete.php
```diff
@@ -0,0 +1,14 @@
...
+ * @subpackage  com_media
+ *
+ * @copyright   Copyright (C) 2005 - 2017 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+defined('_JEXEC') or die;
+$title = JText::_('JTOOLBAR_DELETE');
+?>
+<button class="btn btn-sm btn-outline-danger">
+    <span class="icon- fa fa-trash-o" title="<?php echo $title; ?>"></span> <?php echo $title; ?>
```
yvesh, line 13: same

administrator/components/com_media/resources/scripts/components/app.vue
```diff
@@ -1,45 +1,35 @@
 <template>
-    <div class="media-container" :style="{minHeight: fullHeight}">
-        <media-toolbar></media-toolbar>
-        <div class="media-main">
-            <div class="media-sidebar">
-                <media-tree :root="'/'"></media-tree>
+    <div class="media-container row">
+        <div class="media-sidebar col-md-2 hidden-sm-down">
```
yvesh: sure you want md? In bs4 sm and md switched.

administrator/components/com_media/views/media/view.html.php
```diff
@@ -43,7 +43,49 @@ public function display($tpl = null)
...
+		{
+			// Instantiate a new JLayoutFile instance and render the layout
+			$layout = new JLayoutFile('toolbar.upload', JPATH_COMPONENT_ADMINISTRATOR . '/legacy/layouts');
+
+			$bar->appendButton('Custom', $layout->render(array()), 'upload');
+			JToolbarHelper::divider();
+		}
+
+		// Add an upload button
+		if ($user->authorise('core.create', 'com_media'))
```
yvesh: Why an second if? and not put in the first one?

administrator/components/com_media/layouts/toolbar/create-folder.php
dneukirchen: ```<i>``` isn’t the icon tag, it’s the italic tag, and it doesn’t make much difference which you pick. 
Dont know who the **** came up with the idea to use ```<i>```s, but it is semantically not correct.

```<span>``` is more semantically correct, because it has no semantics.

And regarding icon- class: it is currently needed by the admin theme to render buttons in that style.

administrator/components/com_media/views/media/view.html.php
dneukirchen: Was copy&paste from old mm. Removed with latest commit

Issue #88, opened 2017-02-22
API: error handling / input validation 


dneukirchen:
* Validate / Sanitize input
* Return the correct status / messages codes on errors

laoneo:
Can this be closed with the merge of #104?

dneukirchen:
Yes

PR #104, opened 2017-02-27
Input check and more stable

laoneo:
Pull Request for Issue #88. Upload checks which are applicable with the new architecture are migrated from [old media manager upload function](https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_media/controllers/file.json.php#L29). Any feedback is welcome.

What is not done is the check for `JSession::checkToken('request'))`. Question here, should this be done for every request or only PUT and POST?

### DIFF
--- a/administrator/components/com_media/controllers/api.json.php
+++ b/administrator/components/com_media/controllers/api.json.php
@@ -82,19 +82,18 @@ class MediaControllerApi extends Controller
 		$path = $this->input->getPath('path', '/', 'path');
 
 		// Determine the method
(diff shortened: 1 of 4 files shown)


dneukirchen:
We need the token-check on every endpoint which can modify data. 

That should be POST, PUT and DELETE. 
GET requests should never modify data. So there is no need for the token check on the endpoint.

laoneo:
Closing it as it needs some polish, will reopen when more work is done.

dneukirchen (approved):
lgtm

PR #106, opened 2017-02-28
Feature/add csrf token to client

dneukirchen:
Follow up PR for #104 
Resolves issue #105 

* Add csrf token to ajax requests

### DIFF
--- a/administrator/components/com_media/controllers/api.json.php
+++ b/administrator/components/com_media/controllers/api.json.php
@@ -87,7 +87,7 @@ class MediaControllerApi extends Controller
 		try
 		{
 			// Check token for requests which do modify files (all except get requests)
-			if ($method != 'get' && !JSession::checkToken('request'))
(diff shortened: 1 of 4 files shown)


dgrammatiko (approved):

dgrammatiko:
But CSRF is dead: https://scotthelme.co.uk/csrf-is-dead/ (if only all browsers supported that...)

dneukirchen:
I think SameSite cookies are out of the scope of this project, the browser support ist horrible (http://caniuse.com/#search=SameSite) and i personally do not fully agree with that blog post in the age of stateless rest apis.

dgrammatiko:
Yes, it's way out of the scope of this project, but I won't create another issue that no-one will ever bother to even search a little about what this sameSite is...
Well, depends, the state should live on the client side (at least that's the theory for modern apps) so in that sense the api should be stateless...

@laoneo can you merge this?

Issue #93, opened 2017-02-23
Adjust names of newly created folders


dneukirchen:
Follow up todo from https://github.com/joomla-projects/media-manager-improvement/pull/90

currently all whitespaces are stripped, perhaps it is better to replace them with '-'

laoneo:
Is this an issue on the client side or api?

dneukirchen:
Nope, its handled by the server (JFolder::create()). We have to manually modify it before sending the name to JFolder::create().

laoneo:
Can we close this in favor of #107?

PR #107, opened 2017-03-06
Check the folder name the same way as with files

laoneo:
Pull Request for Issue #93.

### Summary of Changes
Reads the name from the input allowing more characters. The name is sanitized the same way as with file names.

### DIFF
--- a/administrator/components/com_media/controllers/api.json.php
+++ b/administrator/components/com_media/controllers/api.json.php
@@ -103,7 +103,7 @@ class MediaControllerApi extends Controller
 					break;
 				case 'post':
 					$content      = $this->input->json;
-					$name         = $content->get('name');
+					$name         = $content->getString('name');
 					$mediaContent = base64_decode($content->get('content'));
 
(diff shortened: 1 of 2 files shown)


dneukirchen (changes requested):
When i create a new directory with whitespaces in the directory name, im getting a 404 not found error response. The directory is created, but i have to reload the page to see it.

Whitespaces are replaced by underscores not by dashes.

laoneo:
Fixed the 403 error in the last commit.

> Whitespaces are replaced by underscores not by dashes.

That's the [behavior](https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/administrator/components/com_media/models/api.php#L191) from the old media manager to replace the non-alphanumeric characters with an _.

dneukirchen (approved):
Looks good now. Let UX decide on the _ vs - thing.
Can you inform UX about this @laoneo ?

laoneo:
Will do on the meeting today.

PR #108, opened 2017-03-18
Fix incorrect redirect on error handling

jajodiaraghav:
### Summary of Changes
- When an error is encountered, the mediamanager redirects to an incorrect url. Now it redirects back to administrator dashboard.
- The getimagesize() function throws error when a folder is passed as argument. Now it only allows image files to be passed as argument.

### Testing Instructions
Click on media from administrator dashboard, it redirect to "/media-management-improvement/administrator" rather than "/administrator". Also, console error does not occur

None

### DIFF
--- a/administrator/components/com_media/resources/scripts/app/Api.js
+++ b/administrator/components/com_media/resources/scripts/app/Api.js
@@ -113,9 +113,9 @@ class Api {
             case 401:
             case 403:
             case 500:
-                window.location.href = '/administrator';
+                window.location.href = window.location.pathname;
             default:
-                window.location.href = '/administrator';
+                window.location.href = window.location.pathname;
         }
 
         throw error;
--- a/media/com_media/js/mediamanager.js
+++ b/media/com_media/js/mediamanager.js
@@ -410,6 +410,7 @@ process.umask = function() { return 0; };
 
 },{}],3:[function(require,module,exports){
 var Vue // late bind
+var version
 var map = window.__VUE_HOT_MAP__ = Object.create(null)
 var installed = false
 var isBrowserify = false
@@ -419,7 +420,8 @@ exports.install = function (vue, browserify) {
   if (installed) return
   installed = true
 
-  Vue = vue
+  Vue = vue.__esModule ? vue.default : vue
+  version = Vue.version.split('.').map(Number)
   isBrowserify = browserify
 
   // compat with < 2.0.0-alpha.7
@@ -427,7 +429,7 @@ exports.install = function (vue, browserify) {
     initHookName = 'init'
   }
 
-  exports.compatible = Number(Vue.version.split('.')[0]) >= 2
+  exports.compatible = version[0] >= 2
   if (!exports.compatible) {
     console.warn(
       '[HMR] You are using a version of vue-hot-reload-api that is ' +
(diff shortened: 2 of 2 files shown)


laoneo:
Thanks for you contribution. You shoukld not change the mediamanager.js file, instead of you need to set up npm and then change the right file ind the resources directory. Then run `npm install` and `npm run build` in the folder administrator/components/com_media.

jajodiaraghav:
Updated the commits! 
Thanks for the help :)

dneukirchen (changes requested):
This will only work for joomla instances which are installed in a subfolder (my-domain.tld/subfolder/), but will not work for joomla instances on the root level (my-domain.tld) or sub-subfolders (my-
domain.tld/subfolder/another-subfolder/).

administrator/components/com_media/resources/scripts/app/Api.js
```diff
@@ -106,16 +106,19 @@ class Api {
      * @TODO DN improve error handling
      */
     _handleError(error) {
-        alert(error.status + ' ' + error.statusText);
+        if (error.status !== 200) {
+            alert(error.status + ' ' + error.statusText);
+        }
+        var url_array = window.location.href.split('/');
```
dneukirchen: This will only work for joomla instances which are installed in a subfolder (my-domain.tld/subfolder/), but will not work for joomla instances on the root level (my-domain.tld) or sub-subfolders (my-
domain.tld/subfolder/another-subfolder/).

administrator/components/com_media/resources/scripts/app/Api.js
```diff
@@ -106,16 +106,19 @@ class Api {
      * @TODO DN improve error handling
      */
     _handleError(error) {
-        alert(error.status + ' ' + error.statusText);
+        if (error.status !== 200) {
```
dneukirchen: 200 status code responses should never end up in the _handleError method.

dneukirchen (changes requested):

administrator/components/com_media/resources/scripts/app/Api.js
```diff
@@ -106,16 +106,19 @@ class Api {
...
+        }
+        var url_array = window.location.href.split('/');
         switch (error.status) {
             case 404:
                 break;
             case 401:
             case 403:
             case 500:
-                window.location.href = '/administrator';
+                window.location.href = '/' + url_array[3] + '/administrator';
```
dneukirchen: This will only work for joomla instances which are installed in a subfolder (my-domain.tld/subfolder/), but will not work for joomla instances on the root level (my-domain.tld) or sub-subfolders (my-
domain.tld/subfolder/another-subfolder/).

administrator/components/com_media/resources/scripts/app/Api.js
```diff
@@ -106,16 +106,19 @@ class Api {
...
             default:
-                window.location.href = '/administrator';
+                window.location.href = '/' + url_array[3] + '/administrator';
```
dneukirchen: same as above

dneukirchen (changes requested):

libraries/vendor/joomla/image/src/Image.php
```diff
@@ -211,10 +211,13 @@ public static function getImageFileProperties($path)
 			throw new \InvalidArgumentException('The image file does not exist.');
 		}
 
-		// Get the image file information.
-		$info = getimagesize($path);
+		if (!is_dir($path))
```
laoneo: Either way you need to make the pr with the changes in the vendor directory to the upstream package or catch that case in the media manager. Changes here will be overwritten on the next composer update.

jajodiaraghav: I am not very clear about what you mean here.

administrator/components/com_media/resources/scripts/app/Api.js
jajodiaraghav: Well, 200 status code did end up in _handleError and this is due to the `getimagesize($path)` in `libraries/vendor/joomla/image/src/Image.php`. Maybe handling that exception properly might fix this. Any ideas?
Thanks

dneukirchen:
Thank you @jajodiaraghav

PR #116, opened 2017-03-25
Only fetch the width and height when it is a file

laoneo:
When error reporting is set to maximum, then an error is thrown because the image size can't be determined on a folder. This pr fixes that.

The error is:
_Notice:  getimagesize(): Read error! in /mm/libraries/vendor/joomla/image/src/Image.php on line 215_

### DIFF
--- a/administrator/components/com_media/libraries/media/file/adapter/local.php
+++ b/administrator/components/com_media/libraries/media/file/adapter/local.php
@@ -280,7 +280,7 @@ class MediaFileAdapterLocal implements MediaFileAdapterInterface
 		$obj->width                   = 0;
 		$obj->height                  = 0;
 
-		if (strpos($obj->mime_type, 'image/' === 0) && in_array(strtolower($obj->extension), array('jpg', 'jpeg', 'png', 'gif', 'bmp')))
+		if (strpos($obj->mime_type, 'image/') === 0 && in_array(strtolower($obj->extension), array('jpg', 'jpeg', 'png', 'gif', 'bmp')))
 		{
 			// Get the image properties
 			$props       = JImage::getImageFileProperties($path);
--- a/administrator/components/com_media/libraries/media/file/adapter/local.php
+++ b/administrator/components/com_media/libraries/media/file/adapter/local.php
@@ -280,10 +280,16 @@ class MediaFileAdapterLocal implements MediaFileAdapterInterface
 
 		try
 		{
-			// Get the image properties
-			$props       = JImage::getImageFileProperties($path);
-			$obj->width  = $props->width;
-			$obj->height = $props->height;
+			$obj->width  = 0;
+			$obj->height = 0;
+
+			if (!is_dir($path))
+			{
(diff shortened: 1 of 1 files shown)


administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -280,10 +280,16 @@ private function getPathInformation($path)
...
+			$obj->width  = 0;
+			$obj->height = 0;
+
+			if (!is_dir($path))
+			{
+				// Get the image properties
+				$props       = JImage::getImageFileProperties($path);
+				$obj->width  = $props->width;
+				$obj->height = $props->height;
+			}
```
dneukirchen: This will fix the error for directories but what happens in case of non-image files like html, mp4, pdf? Perhaps we need an isImage() check.

laoneo:
Then an exception is thrown, which we catch.

dneukirchen:
Hm, is that what we want? Imho exceptions should be thrown (and catched) only in case of serious errors (when the app is unable to recover) and not as part of an expected and valid application flow.

administrator/components/com_media/libraries/media/file/adapter/local.php
laoneo: Most information you find on the net do say that getimagesize is the function to check if it is an image or not. That is what the JImage lib is doing. I'm not sure if it would be enough to rely on the filetyope or mime type only.

administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -277,20 +277,16 @@ private function getPathInformation($path)
 		$obj->modified_date           = $modifiedDate->format('c', true);
 		$obj->modified_date_formatted = (string) $modifiedDate; // TODO use format from config
 		$obj->mime_type               = mime_content_type($path);
+		$obj->width                   = 0;
+		$obj->height                  = 0;
 
-		try
+		if (strpos($obj->mime_type, 'image/') === 0 && in_array(strtolower($obj->extension), array('jpg', 'jpeg', 'png', 'gif', 'bmp')))
```
dneukirchen, line 283: Can we put this in an isImage method to be more readable?

administrator/components/com_media/libraries/media/file/adapter/local.php
```diff
@@ -277,20 +277,16 @@ private function getPathInformation($path)
...
 		{
 			// Get the image properties
 			$props       = JImage::getImageFileProperties($path);
 			$obj->width  = $props->width;
 			$obj->height = $props->height;
 		}
-		catch (Exception $e)
```
dneukirchen, line 288: What kind of exception do we catch here?

laoneo: I'v removed that and opened issue #117 as we don't have a defined way of exception handling.

dneukirchen:
thanks @laoneo

PR #122, opened 2017-03-29
An image edit view with action plugins, based on JForm

laoneo:
This pr adds a first approach of an image edit view which adds image manipulating functionality trough plugins. Each plugin can add fieldsets to a JForm instance which will be then rendered as tabs in the view.

image

An event bus supports to add event listeners by the plugins. The `onActivate` event is triggered when a tab is activated. The `onDeactivate` is triggered for an active tab which got deactivated.

There is a base `MediaActionPlugin`class which supports out of the box to load a form file into the JForm instance and to load a layout to add the JS code to the event bus.

### Todo
There's still stuff left after this pr:
- There are probably better JS libraries out there, some research should be done which one to use.
- Save actions do not work.
- Cancel action does not work.
- Initial plugin activate on page load is not fired.


media/com_media/js/EventBus.js
```diff
@@ -0,0 +1,100 @@
...
+	else
+		root["EventBus"] = factory();
+})(this, function() {
+
+	var EventBusClass = {};
+	EventBusClass = function() {
+		this.listeners = {};
+	};
+	EventBusClass.prototype = {
+		addEventListener:function(type, callback, scope) {
```
dgrammatiko, line 10: Either import the library as is or remove the amd, requirejs part. I guess the first, although I still don't see the need here

laoneo: No need for what?

dgrammatiko:
```js
(function (root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define("EventBus", [], factory);
	else if(typeof exports === 'object')
		exports["EventBus"] = factory();
	else
		root["EventBus"] = factory();
})(this, function() {
```
is useless as we don't use amd or requirejs

laoneo: Will do, understood it the way that the whole event bus doesn't make sense.

N6REJ:
instead of reinventing a image editor, why are we not using a well-established/tested one..  Joomla has had a photo-editor extension for close to a decade now, and there are VERY well established editors that rival desktop editors like pixlr

dgrammatiko:
@N6REJ no way we'll use code that's build upon flash:
screen shot 2017-03-29 at 22 38 09

media/com_media/js/EventBus.js
dgrammatiko: That one and also instead of the event bus you can create Joomla.mediaBus and do your magic there. From my point of view this is just extra clutter

N6REJ:
didn't say it has to be flash, ijs there are already really high quality processor in the FOSS world.  @dgt41 I think you know this.

N6REJ:
we could always use gimp LOL

N6REJ:
Here's one for example... [photo Editor](http://camanjs.com/examples/)  and we could use [live Sliders](http://www.virtuosoft.eu/code/jquery-colorpickersliders/) to make it even better

laoneo:
That's why we made the plugin based approach, you can then make the all in one plugin and you'r fine. We don't want to stick to one specific implementation.

dneukirchen:
thank you @laoneo 
We will work together on the details and refactor some parts of your approach.

Issue #121, opened 2017-03-27
Media Manager shows a empty pane


kasvith:
### Steps to reproduce the issue
- Go to admin panel
- Go to media manager

### Expected result
- A hierarchy with contents uploaded into media storage 

### Actual result
- Empty pane

laoneo:
Can you post a acreenshot where you get the empty pane?

kasvith:
![screenshot_131](https://cloud.githubusercontent.com/assets/13379595/24365714/7fca6c1e-1334-11e7-8d3c-2cc342a66ad6.png)

Content folder shown to the right

dneukirchen:
Please open webdeveloper toolbar and look in console and network tab. Any errors there?

kasvith:
yes there is
 Console Tab
> error TypeError: Cannot read property 'directories' of undefined
    at http://localhost/joomla4/media/com_media/js/mediamanager.js:9253:70
    at wrappedMutationHandler (http://localhost/joomla4/media/com_media/js/mediamanager.js:7936:5)
    at commitIterator (http://localhost/joomla4/media/com_media/js/mediamanager.js:7660:7)
    at Array.forEach (native)
    at http://localhost/joomla4/media/com_media/js/mediamanager.js:7659:11
    at Store._withCommit (http://localhost/joomla4/media/com_media/js/mediamanager.js:7747:3)
    at Store.commit (http://localhost/joomla4/media/com_media/js/mediamanager.js:7658:8)
    at boundCommit (http://localhost/joomla4/media/com_media/js/mediamanager.js:7614:19)
    at http://localhost/joomla4/media/com_media/js/mediamanager.js:9095:9

dneukirchen:
In the network tab you should see a xhr request to our api endpoint (index.php?option=com_media...). 

Does it return a 200 OK status? Can you post the response content here?

Thanks.

kasvith:
@dneukirchen yes there it is with HTTP 200 OK

dneukirchen:
Can you post the response here?

kasvith:
[...]

PR #124, opened 2017-03-30
Use OS path separator

laoneo:
Pull Request for Issue #121.

Uses the OS directory path separator instead of /.


mbabker:
You could make use of `JPath::clean()` here.  It also deals with standardizing separators.

ciar4n:
This doesn't appear to resolve the issue on #121

N6REJ:
I have to agree with Michael.. JPath::clean() is the right way to go.

Issue #126, opened 2017-04-01
Move Edit view layouts to files


laoneo:
Actually the plugin JS code of the media action plugins is in a layout, it needs to be moved to a JS file and loaded automatically by the base media action class.

PR #137, opened 2017-04-01
Move js code to files in media action plugins

laoneo:
Pull Request for Issue #126.

### Summary of Changes
Moves the js code from the layouts to js files. Additionally the lib files are stored locally with a minified and a dev version.


dgrammatiko (approved):
I like the simplicity here

administrator/components/com_media/views/file/tmpl/default.php
```diff
@@ -13,8 +13,8 @@
 JHtml::_('jquery.framework');
```
dgrammatiko, line 13: we should remove that line

laoneo:
Hopefully it will make it easier for the JS cracks to work on the media action plugins :wink:

administrator/components/com_media/views/file/tmpl/default.php
laoneo: It's still needed because the edit.js file is jQuery dependent. If somebody with more knowledge can fix that, would be nice to remove it.

dgrammatiko: what the heck is edit.js?

dgrammatiko: Ok replace the code with:
```js
(function() {
  "use strict";
	document.addEventListener('DOMContentLoaded', function() {
		var list = document.querySelectorAll('a[data-toggle="tab"]');

		list.forEach(function(item) {
			item.addEventListener('shown.bs.tab', function(event) {
				if (event.relatedTarget) {
					EventBus.dispatch('onDeactivate', this, event.relatedTarget.hash.replace('#attrib-', ''), document.getElementById('media-edit-file'));
				}
				EventBus.dispatch('onActivate', this, event.target.hash.replace('#attrib-', ''), document.getElementById('media-edit-file'));
			})
		})
	});
})();
```

Issue #130, opened 2017-04-01
Namespace event bus


laoneo:
The event bus should be namespaced to Joomla.Media.EventBus.

dgrammatiko:
I still think you should drop it as it's not essential for our use case

dneukirchen:
Closing this. Eventbus was removed in https://github.com/joomla-projects/media-manager-improvement/pull/152/files

PR #138, opened 2017-04-01
[WIP] Make the cropper plugin properly work

dgrammatiko:
Pull Request for Issue # .


laoneo:
Did you test the resize, crop and rotate plugins? Are they still working?

dneukirchen (commented):
I dont really like putting component specific deps in the media vendors folder, but im ok with it in this case.

Perhaps it would be a good idea to namespace it with Joomla.Event? 

Would be also nice to wrap the specific implementation (eventbausjs) in a joomla wrapper, so that we can exchange it when needed.  In that wrapper we could also simplify the naming:

* `fire` instead of `dispatch`
* `listen` instead of `addEventListener`

So it comes down to `Joomla.Event.fire(event, payload)` and `Joomla.Event.listen(event, callback)`.

dneukirchen:
#130

laoneo:
@dgt41 what do you think to split that pr into multiple ones? It is hard to review at the moment as too many things are done here at the same time.

dneukirchen:
@dgt41 should we close this one in favor of #152 and upcoming prs?

Issue #114, opened 2017-03-23
Submit create-folder-modal-form with enter key


dneukirchen:
### Steps to reproduce the issue

Open create folder modal.
Fill in the folder name.
Hit enter key.

### Expected result

New folder should be created
Modal should be closed

dneukirchen:
@kasvith will resolve this one.

kasvith:
sure in ASAP :)

dneukirchen:
thanks @kasvith

PR #147, opened 2017-04-02
Enhancement/create folder keypress close

kasvith:
Pull Request for Issue #114  .

### Summary of Changes
- Added functionality to create folder with Enter key press
- Shows an error to user if tried to create a folder with empty name

### Expected result
- Create new folder with enter keypress.

### Actual result
- Create new folder with enter keypress.

- Do changes to necessary translations


administrator/components/com_media/package.json
```diff
@@ -1,31 +1,31 @@
...
+		"node-sass": "^4.5.1",
+		"vueify": "^9.4.1"
+	},
+	"browserify": {
+		"transform": [
+			"vueify",
+			"babelify"
+		]
+	}
 }
```
dneukirchen: Code style. Please change indentation.

administrator/components/com_media/resources/scripts/components/modals/create-folder-modal.vue
```diff
@@ -23,13 +23,30 @@
...
         methods: {
             /* Close the modal instance */
             close() {
                 this.$store.commit(types.HIDE_CREATE_FOLDER_MODAL);
             },
             /* Save the form and create the folder */
             save() {
+                /*Checks for the folder name, if it is not declared display an error to user and abort task*/
+                if(!this.folder){
+                    this.$store.commit(types.HIDE_CREATE_FOLDER_MODAL);
```
dneukirchen: Hm. Do we really want to close the modal when an validation error occurs? 

If the user submits the form, he most probably dont want the modal to be closed in case the form is invalid. 

I believe you close the modal because the error message is behind the modal? Imo this is a css issue which should be fixed or solved in a different way. Closing the modal is not very user friendly.

Perhaps we can use the scd argument of Joomla.renderMessages (The selector of the container where the message will be rendered). 

Additionally we need to mark the input field as invalid (red border, label).

kasvith:
@dneukirchen does it work ?

administrator/components/com_media/resources/scripts/components/modals/create-folder-modal.vue
```diff
@@ -23,13 +23,30 @@
...
         directives: { focus: focus },
+        ready: function() {
+            /*add event listener on keydown*/
+            document.addEventListener("keydown", (e) => {
+                /*checks if the model is shown and enter(keycode 13) is pressed*/
+                if($store.state.showCreateFolderModal && e.keyCode == 13){
+                    save();
+                    close();
+                }
+            });
```
dneukirchen: Do we need this additional event listener? @submit.prevent="save" should do the trick without it.

dneukirchen (changes requested):

administrator/components/com_media/resources/scripts/components/modals/create-folder-modal.vue
kasvith: Yes it was the reason, if we can fix css issue i will not close it

administrator/components/com_media/package.json
kasvith: Ohhh again :/

administrator/components/com_media/resources/scripts/components/modals/create-folder-modal.vue
kasvith: Will try that

dneukirchen (approved):

dneukirchen:
thanks @kasvith

kasvith:
😊

PR #157, opened 2017-04-05
Fix upload content

laoneo:
The content in PUT and POST is not properly retrieved from the input.


dneukirchen (approved):

dneukirchen:
thanks @laoneo.

We can ignore drone failure.

PR #160, opened 2017-04-05
Basic file upload

dneukirchen:
This PR adds the basic file upload feature. 

# Testing instructions

Click upload button in toolbar
Select one or more files
See the file(s) uploaded in the current selected folder.

# Todo
There is a lot to do until we have a final version of this feature:

* Show Progress
* Handle errors (filesize, file already exists...)
* Show Success/Error notifications
* Limit upload on client side (just for convenience. security is handled on server side)
* Check security on server side 
* Add correct html attributes to file input
...

I will work on some of the todos and create issues for them after this one is merged.


dgrammatiko (approved):

dneukirchen:
Drone is failing because of a bug in js tests. Its already fixed in main repo: https://github.com/joomla/joomla-cms/pull/14922

administrator/components/com_media/resources/scripts/components/upload/upload.vue
```diff
@@ -0,0 +1,67 @@
...
+                // Loop through array of files and upload each file
+                for (let file of files) {
+
+                    // Create a new file reader instance
+                    let reader = new FileReader();
+
+                    // Add the on load callback
+                    reader.onload = (progressEvent) => {
+                        const result = progressEvent.target.result,
+                            splitIndex = result.indexOf('base64') + 7,
```
dgrammatiko, line 47: does this cover jpeg and webp (these are 4 char long)?

dneukirchen: yes

dgrammatiko: Great! 👍

dneukirchen: I was not 100% sure about this part... It is something that could be easily done on server-side: check if content starts with ... and remove it. But i decided to put in in the client until we have a good reason to not do so.

administrator/components/com_media/resources/scripts/components/browser/items/item.js
```diff
@@ -16,7 +16,7 @@ export default {
          * Return the correct item type component
          */
         function itemType() {
-            let imageExtensions = ['jpg', 'png', 'gif'];
+            let imageExtensions = ['jpg', 'jpeg', 'png', 'gif'];
```
laoneo, line 19: Should we support bmp images as well?

dneukirchen: You are right. This needs to be refactored to be flexible and extendable, but it is imo out of the scope of this PR.

laoneo:
Nice work David!

N6REJ:
shouldn't it be following what is permitted in J!?  why are we having to recreate so much?

dneukirchen:
@N6REJ i think this is a misunderstanding. 

For the the question "what extensions are allowed to upload?" you are right and we are using the values from the joomla config there, but the quoted lines of code are not related to upload restrictions. 

In the itemType() method we decide which vue component to render in the media grid view. We are using different components for images, folders and files, because they have different attributes and behaviour (i.e. if you double click on folder -> you navigate to that folder. If you double click on image -> that will open the image preview/edit). 

This will be extendable by plugins in the future, so that you can register your own media type like svg, psd, mp3, .csv, google docs or whatever.

PR #170, opened 2017-04-07
Initial folder tree styling

dgrammatiko:
Pull Request for Issue # .

PR #171, opened 2017-04-07
Remove level property

dneukirchen:
Remove the level property.

see #170


ciar4n:
Thank you @dneukirchen . I missed that.

dgrammatiko (approved):

Issue #159, opened 2017-04-05
Move local adapter to Joomla plugin


laoneo:
Actually the adapter which implements the `MediaFileAdapterInterface` is injected into the [constructor of the api model](https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/administrator/components/com_media/models/api.php#L39).

To be prepared for multiple adapters in a model, the local adapter should be put into a Joomla plugin and then loaded trough Joomla events in the model.

kasvith:
@laoneo is this plugin supposed to be used as the base plugin, for 3rd party media manager extensions ?
So they can use this for save files from clouds to local

laoneo:
For this issue it is not relevant if the local plugin will serve as base plugin for the cloud plugins or not. This question we have to answer then when we start to implement the cloud adapters, but for now you can concentrate on the move only.

laoneo:
Pr is merged.

PR #183, opened 2017-04-12
FileSystemLocal adapter plugin

kasvith:
Pull Request for Issue #159  .

### Summary of Changes
- Moved local adapter from **com_media** to a plugin  **FileSystemLocal** 

### Testing Instructions
- Discover and Enable plugin if it is not enabled yet
- Open Media Manager

### Expected result
- Working Media Manager with new Local FileSystem plugin

### Actual result
- Working local filesystem with new plugin


administrator/components/com_content/dispatcher.php
```diff
@@ -28,7 +28,8 @@ class ContentDispatcher extends Dispatcher
 	 *
 	 * @since   __DEPLOY_VERSION__
 	 */
-	public function __construct($namespace, \Joomla\CMS\Application\CmsApplication $app, \JInput $input = null, \Joomla\CMS\Mvc\Factory\MvcFactoryInterface $factory = null)
+	public function __construct($namespace, \Joomla\CMS\Application\CmsApplication $app, \JInput $input = null,
```
laoneo: Please revert that change.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -81,6 +84,98 @@ public function getFile($path = '/')
 	}
 
 	/**
+	 * Returns the folder or file information for the given path. The returned object
```
laoneo: Why did you change the whole block here?

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -12,6 +12,9 @@
 JLoader::import('joomla.filesystem.file');
 JLoader::import('joomla.filesystem.folder');
 
+require_once 'filenotfoundexception.php';
```
laoneo: These files need to stay in com_media as they are the same for all future cloud plugins.

laoneo (changes requested):
How does the api model get then the adapter at all?

laoneo:
A good start, but needs some changes.

plugins/filesystem/local/adapter/adapter.php
kasvith: what you meant ?

plugins/filesystem/local/adapter/adapter.php
kasvith: is there a joomla way to import them ?

administrator/components/com_content/dispatcher.php
kasvith: Done, did due to build fail

plugins/filesystem/local/adapter/adapter.php
kasvith: or can i use require_once for that, if it does not violating Joomla! rules

kasvith: will move it to com_media and import to this file

kasvith: Anyway if these are common to cloud filesystems
Why we don't use this plugin as the one who really deals with files, like PlgFileSystemGoogleDrive does not need to know how a file handled in Local File System, it is encapsulated on this plugin, it can provide the functionality without a change.

Any comments about that approach ?
I moved those keeping that design in my mind

plugins/filesystem/local/adapter/adapter.php
kasvith: There were some code syntax issues in this file, in incorrect format. I fixed them to Joomla! standard

there was like

if(){

}

statements and changed them to

if()
{

}

kasvith:
@laoneo oops sorry after a pull request my code has been changed, thanks for mentioning it above 👍

plugins/filesystem/local/adapter/adapter.php
laoneo: But the api model needs to know what it gets when it asks for an adapter, so the interface needs to stay there. com_media should not have any dependency to a plugin but the plugins to com_media.

plugins/filesystem/local/adapter/adapter.php
laoneo: If you look on the code removed and added, it's the same.

plugins/filesystem/local/adapter/adapter.php
kasvith: will fix that ASAP

plugins/filesystem/local/adapter/adapter.php
kasvith: oh i see, i think i've used space before, then changed back to tabs
Can you mark it as resolved :)

kasvith:
@laoneo can you check now ?

laoneo:
Please move the interfaces back to their original place. Thanks.

kasvith:
Moved interfaces to *com_media/libraries/media/file/adapter*

plugins/filesystem/local/adapter/adapter.php
laoneo: Did you modify something in that code? There is for me no reason why it should be changed. For me it looks identical.

laoneo (changes requested):

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -12,6 +12,13 @@
 JLoader::import('joomla.filesystem.file');
 JLoader::import('joomla.filesystem.folder');
 
+// Path for Interfaces in com_media
+define('MEDIA_INTERFACE_PATH', JPATH_ADMINISTRATOR . '/components/com_media/libraries/media/file/adapter/');
+
+include_once MEDIA_INTERFACE_PATH . 'interface.php';
```
laoneo: They should be loaded trough JLoader like here https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/plugins/media-action/crop/crop.php#L12

kasvith:
@laoneo it is changing **spaces** to **tabs** caused that problem

laoneo:
The original file doesn't have spaces.

kasvith:
yes, but when i imported file to PHPStrom it asked me to switch to tabs, so i switched at first time and commited

plugins/filesystem/local/adapter/adapter.php
kasvith: I first commited to my repo with spaces, may that be the cause ?

kasvith:
Used **JLoader**

plugins/filesystem/local/adapter/adapter.php
laoneo: Just copy it from the original file and then the changes should disappear.

plugins/filesystem/local/local.php
```diff
@@ -0,0 +1,45 @@
...
+ * @package     Joomla.Plugin
+ * @subpackage  FileSystem.Local
+ *
+ * @copyright   Copyright (C) 2005 - 2017 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('_JEXEC') or die;
+
+require_once 'adapter/adapter.php';
```
laoneo: Here also JLoader

administrator/components/com_media/models/api.php
```diff
@@ -9,6 +9,7 @@
 
 defined('_JEXEC') or die;
 
+
```
laoneo: This one here is not needed

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -11,7 +11,8 @@
 
 JLoader::import('joomla.filesystem.file');
 JLoader::import('joomla.filesystem.folder');
-
+JLoader::import('components.com_media.libraries.media.file.adapter.interface', JPATH_ADMINISTRATOR);
+JLoader::import('components.com_media.libraries.media.file.adapter.filenotfoundexception', JPATH_ADMINISTRATOR);
 /**
```
laoneo: Here is a space needed.

laoneo (changes requested):

plugins/filesystem/local/local.php
kasvith: with same method ?

laoneo: Yes

kasvith: done

administrator/components/com_media/models/api.php
kasvith: using CMS over cms ?

laoneo:
Thank you @kasvith, good work!

Issue #186, opened 2017-04-13
File/Folder Copy, Move support in the local adapter


kasvith:
It would be nice to implement the File/Folder 

- Copy
- Move
- Rename

functionalities in the backend local adapter, so frontend developers can use them to make basic file operations work

kasvith:
# Directory copy/move when a directory with same name exists on the destination

## Solution 01
In this scenario we can decide to copy non existing files individually first to the existing directory
**If there are files with same name on the target directory too** user may want to **overwrite** or **skip** these files.

For that frontend should know what files are being overwritten

For overcome this, we can put all files that are about to be overwritten in a queue and send it to the frontend, so user can decide each file and backend can process files with given information

## Solution 02
In the same scenario, instead of sending the *copy, move, rename* operations directly to the adapter, front end can check for availability, if files with same name exists on the destination dir, frontend can warn the user and if he wants to overwrite them, he can do it in frontend, then a request sent to the backend with a **force** condition, if **force** is **true** backend will simply overwrite any file silently.

This requires to load the folder structure inside the filesystem when frontend loads, so analysis can be done in frontend and final action could be taken at the backend

Any comments @laoneo @dneukirchen @dgt41

laoneo:
I would go for solution 2 as the api should not become stateful to manage some queue for the next request from the client. This is better managed in the client.

kasvith:
And also it is easy for implement, less load on the backend

dgrammatiko:
Could we utilise promises on the client side and adjust the backend for that?
The code will be much simpler...

kasvith:
Can you give an example for that

dgrammatiko:
Create folder (as xhr) returns promise, if ok folder created else the folder exists and an alert is thrown with the ability to rename, overwrite or cancel

kasvith:
Yes we can do as you said for the frontend, in here i was bit focused on backend
For backend functions will return a **true** on **success**, will throw exceptions at any failure

kasvith:
# Mechanism for file/folder Copy/Move

- Check if the target directory of file **exists**. 
  - If it does not exists, use the provided method on adapter to copy/move. If it is a folder, all the contents will be copied to the target directory. If it is a file target will have a copy of file.

- If it exists, we need to handover control to the user, whether files with identical names overwrite or skipped(Folder merging)
  - If this is a file, do a simple request to backend with *force* condition.
  - If this is a directory things will be little bit complex.
[...]

PR #188, opened 2017-04-15
Enhancement/basic file operations in FileSystemLocal Plugin

kasvith:
Pull Request for Issue #186 .

### Summary of Changes
Added following to **FileSystemLocal** plugin
- Move
- Copy

Added Tests for *Move/Copy*

### Testing Instructions
- test using `libraries\vendor\bin\phpunit --testsuite plugins`

### Expected result
- Should be able copy and move files with FileSystemLocal plugin
- Should all test cases pass

### Actual result
- Tests are passing


kasvith:
Please merge #159 before this one :)

laoneo:
We are first finalizing #183.

laoneo:
This pr is empty, guess you messed something up with the merge of milestone1.

kasvith:
wait i will check

kasvith:
It seems to be

laoneo:
Unit tests are missing, then I think it will be ready to merge.

kasvith:
@laoneo I am writing some UnitTests. You can merge it ASAP ⚡️

kasvith:
@laoneo @dneukirchen I have added Unit Tests
Hope it is ready to merge

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
```diff
@@ -359,4 +359,131 @@ public function testDeleteInvalidPath()
 		// Fetch the file from the root folder
 		$adapter->delete('invalid');
 	}
+
+	public function testCopy()
```
laoneo: Please split individual tests into their own function. Every test function should do only one test like copy file or copy folder.

dneukirchen (commented):
I have a few issues with this PR:

1) Interface Extension
We should extend the interface with the new methods: https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/administrator/components/com_media/libraries/media/file/adapter/interface.php

2) Code style 
Im not the biggest friend of multiple nested if/else statements and i think the code can be simplified a lot. 
You can remove most of the else blocks and use early returns to make the code more readable and extract some of the code to protected methods (like moveFile(), moveFolder(), copyFile(), copyFolder()) 
Not a problem, but nice to have: we should unify the file-or-folder-detection in the adapter (sometimes we use is_file, sometimes is_dir).

kasvith:
I thought that to separate methods too, copyFile, copyFolder etc. I will fix that and that will resolve the issue with test cases too

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
kasvith: Will separate methods instead and check them, i wrote methods to handle anykind of copying folders and files, but it seems better to separate them for better readability

kasvith:
Interface will be extended after finializing this one

laoneo:
To the outside world only copy and move should be public, the other functions like copyFile should be private. Don't want to bloat the interface.

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
laoneo: Thats a principle of unit testing.

kasvith:
@dneukirchen what you meant by this ?
>  but nice to have: we should unify the file-or-folder-detection in the adapter (sometimes we use is_file, sometimes is_dir).

kasvith:
# Changes made

- Used **moveFile, moveFolder** functions for better readiabilty
- Used **copyFile, copyFolder** functions for better readiabilty

Updated test cases for above functions

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+	 * @param   string  $sourcePath       Source path of the file or directory
+	 * @param   string  $destinationPath  Destination path of the file or directory
+	 * @param   bool    $force            Set true to overwrite files or directories
+	 *
+	 * @return bool
+	 *
+	 * @since __DEPLOY_VERSION__
+	 * @throws MediaFileAdapterFilenotfoundexception
+	 */
+	public function copy($sourcePath, $destinationPath, $force = false)
```
laoneo, line 344: If you default force to false here, then you need to do that in the interface as well.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+	 *
+	 * @param   string  $sourcePath       Source path of the file or directory
+	 * @param   string  $destinationPath  Destination path of the file or directory
+	 * @param   bool    $force            Set true to overwrite files or directories
+	 *
+	 * @return bool
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected function copyFile($sourcePath, $destinationPath, $force = false)
```
laoneo: Set it to private, when there is a need we can always set it to protected. Going back will be harder.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+	 *
+	 * @param   string  $sourcePath       Source path of the file or directory
+	 * @param   string  $destinationPath  Destination path of the file or directory
+	 * @param   bool    $force            Set true to overwrite files or directories
+	 *
+	 * @return bool
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected function copyFolder($sourcePath, $destinationPath, $force = false)
```
laoneo: Same here as for copyFile.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+	 * @param   string  $sourcePath       Source path of the file or directory
+	 * @param   string  $destinationPath  Destination path of the file or directory
+	 * @param   bool    $force            Set true to overwrite files or directories
+	 *
+	 * @return bool
+	 *
+	 * @since __DEPLOY_VERSION__
+	 * @throws MediaFileAdapterFilenotfoundexception
+	 */
+	public function move($sourcePath, $destinationPath, $force = false)
```
laoneo, line 469: Same here as for copy, default force to false in the interface.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+	{
+		if (file_exists($destinationPath))
+		{
+			if (is_dir($destinationPath))
+			{
+				// We need to bypass exception thrown in JFolder when destination exists
+				// So we only copy it in forced condition, then delete the source to simulate a move
+				if ($force)
+				{
+					$copy_success = JFolder::copy($sourcePath, $destinationPath, '', $force);
```
laoneo: Names must be camel case.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,272 @@ private function getDate($date = null)
...
+				if ($force)
+				{
+					$delete_success = JFile::delete($destinationPath);
+					$move_success = false;
+					if ($delete_success)
+					{
+						$move_success = JFolder::move($sourcePath, $destinationPath);
+					}
+
+					return self::convertToBool($move_success) && $delete_success;
```
laoneo: IS this here really needed convertToBool?

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
```diff
@@ -359,4 +359,187 @@ public function testDeleteInvalidPath()
 		// Fetch the file from the root folder
 		$adapter->delete('invalid');
 	}
+
+	public function testCopyFile()
```
laoneo: Split here the tests like with/without force or with invalid parameters.

laoneo (changes requested):

plugins/filesystem/local/adapter/adapter.php
kasvith: Yes, as in move it returns a error text when failing

plugins/filesystem/local/adapter/adapter.php
kasvith: sure

plugins/filesystem/local/adapter/adapter.php
kasvith: will do that

plugins/filesystem/local/adapter/adapter.php
kasvith: +1

plugins/filesystem/local/adapter/adapter.php
kasvith: +1

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
kasvith: you mean different methods ?

kasvith: like `testFileCopyForce()` etc ?

plugins/filesystem/local/adapter/adapter.php
kasvith: I used this function because of this in JFolder move method
It returns a mixed value when it fails, for this method we want to know whether it was **success** or not
         

> /**
> 	 * Moves a folder.
> 	 *
> 	 * @param   string   $src          The path to the source folder.
> 	 * @param   string   $dest         The path to the destination folder.
> 	 * @param   string   $path         An optional base path to prefix to the file names.
> 	 * @param   boolean  $use_streams  Optionally use streams.
> 	 *
> 	 * @return  mixed  Error message on false or boolean true on success.
> 	 *
> 	 * @since   11.1
> 	 */

https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/libraries/joomla/filesystem/folder.php

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
laoneo: Yes exactly.

plugins/filesystem/local/adapter/adapter.php
laoneo: Them we do it like in the other functions like delete. Return type is void and when something goes wrong an exception will be thrown.

kasvith: I looked at how delete implemented in the adapter, actually the JFolder::delete returns a boolean, true or false on success, in our case move returns a true or mixed error message, in that case we want to check whether it was a boolean to perform, error message will be also taken as true in php

laoneo: If an error message is returned, you need to throw an exception.

plugins/filesystem/local/adapter/adapter.php
kasvith: done

plugins/filesystem/local/adapter/adapter.php
kasvith: done

kasvith:
# Summery of changes
- Separated tests for each case
- Removed boolean return from **copy,move** and made them return a void
- Improved move method

@laoneo @dneukirchen

administrator/components/com_media/libraries/media/file/adapter/interface.php
```diff
@@ -119,4 +119,32 @@ public function updateFile($name, $path, $data);
...
 	public function delete($path);
+
+	/**
+	 * Moves a file or folder from source to destination
+	 *
+	 * @param   string  $sourcePath       The source path
+	 * @param   string  $destinationPath  The destination path
+	 * @param   bool    $force            Force to overwrite
+	 *
+	 * @return bool
```
laoneo: Must be void or?

administrator/components/com_media/libraries/media/file/adapter/interface.php
```diff
@@ -119,4 +119,32 @@ public function updateFile($name, $path, $data);
...
+	public function move($sourcePath, $destinationPath, $force = false);
+
+	/**
+	 * Copies a file or folder from source to destination
+	 *
+	 * @param   string  $sourcePath       The source path
+	 * @param   string  $destinationPath  The destination path
+	 * @param   bool    $force            Force to overwrite
+	 *
+	 * @return bool
```
laoneo: Must be void or?

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,283 @@ private function getDate($date = null)
...
+	/**
+	 * Copies a file or folder to a destination
+	 * If the destination folder or file already exists, it will not overwrite them without
+	 * force.
+	 *
+	 * @param   string  $sourcePath       Source path of the file or directory
+	 * @param   string  $destinationPath  Destination path of the file or directory
+	 * @param   bool    $force            Set true to overwrite files or directories
+	 *
+	 * @return bool
```
laoneo: Must be void or?

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,283 @@ private function getDate($date = null)
...
+			{
+				$success = JFile::copy($sourcePath, $destinationPath);
+			}
+		}
+
+		if (!$success)
+		{
+			throw new Exception('Copy is not possible');
+		}
+
```
laoneo: Remove empty line

tests/unit/suites/plugins/filesystem/local/adapter/MediaFileAdapterLocalTest.php
```diff
@@ -359,4 +359,309 @@ public function testDeleteInvalidPath()
...
+
+    /**
+     * Cleans the root folder
+     *
+     */
+    private function cleanRootFolder()
+    {
+        JFolder::delete($this->root);
+        JFolder::create($this->root);
+	}
```
laoneo: Wrong intend, also on the functions below

laoneo (changes requested):

plugins/filesystem/local/adapter/adapter.php
kasvith: sorry fixed it

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,283 @@ private function getDate($date = null)
...
+					if ($deleteSuccess)
+					{
+						$value = JFolder::move($sourcePath, $destinationPath);
+					}
+				}
+
+				if ($value === false)
+				{
+					throw new Exception($value);
```
laoneo: Why do you throw here an exception with false and no string?

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,283 @@ private function getDate($date = null)
...
+
+				$success = false;
+				if ($force)
+				{
+					$copySuccess = JFolder::copy($sourcePath, $destinationPath, '', true);
+					$deleteSuccess = false;
+
+					if ($copySuccess)
+					{
+						$deleteSuccess = JFolder::delete($sourcePath);
```
laoneo, line 552: Should we not merge the folders?

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,283 @@ private function getDate($date = null)
...
+			}
+		}
+		else
+		{
+			// Perform usual move
+			$value = JFolder::move($sourcePath, $destinationPath);
+
+			if ($value === false)
+			{
+				throw new Exception($value);
```
laoneo: Why do you throw here an exception with false and no string?

plugins/filesystem/local/adapter/adapter.php
kasvith: we cant use move to merge folders, so first we copy and it will merge, then we delete source, php rename does not merge folders

plugins/filesystem/local/adapter/adapter.php
kasvith: sorry it was a mistake, did not notice that, updated file

plugins/filesystem/local/adapter/adapter.php
laoneo: It's not fixed.

plugins/filesystem/local/adapter/adapter.php
```diff
@@ -326,4 +326,278 @@ private function getDate($date = null)
...
+					else
+					{
+						// Undo previous copy
+						JFolder::delete($destinationPath);
+					}
+
+					$success = $copySuccess && $deleteSuccess;
+				}
+
+				if (!$success)
```
laoneo, line 563: Sometimes you do $value === false and sometimes !$success, can this be unified in some way?

kasvith: in value === false we check for explicit false for the statement, so when a text was returned we can distinguish, i can change !success to sucsess === false to unify if wanted

laoneo: So what we do when a text is returned?

kasvith: we throw it as a error

laoneo: Where?

kasvith: ah wait something has been messed up, i'll update

kasvith: fixed it by replacing === false with !== true

kasvith: move returns a true on success and text on error, bit messed it up

laoneo (approved):

laoneo:
Thanks @kasvith!

kasvith:
Thank you @laoneo and @dneukirchen for reviews

PR #200, opened 2017-04-29
Simplify logic in local adapter

laoneo:
This is a follow up PR of #188 which simplifies the logic in the move and copy functions of the great work from @kasvith.


kasvith:
@laoneo thanks

kasvith:
@laoneo does all the test cases passes ?
If they pass it may be ready to merge, i'll give a check

laoneo:
Yes. Guess it just needs a merge by somebody.

Issue #127, opened 2017-04-01
Style the edit view


laoneo:
The actual edit view is very simple, despite that it needs some styling.

dgrammatiko:
@ciar4n this one is for you

ciar4n:
Any preference in how the edit view should be laid out?

laoneo:
Not from my side. Let your fantasy flow :-)

ciar4n:
That could be dangerous :-)

I'll come up with something

kasvith:
Make it something like a basic paint

dneukirchen:
thx @ciar4n

PR #206, opened 2017-05-11
Column layout for edit view

ciar4n:
Pull Request for Issue #127 .

### Summary of Changes
A layout change to the edit view. Not sure what else I can do here as I can't edit the markup as easily as the browse page.


laoneo:
I'm not sure if we should do three columns. If yes, then the plugins area should align them vertically like an accordion or not? When there are many plugins activated the tabs get stacked then?
Also some margins between the edit area and the preview would be nice to have.
image

ciar4n:
Do we need the preview? The edit and preview pretty much display the same thing. Ideally I think plugin options and image should be visible without scroll.

How many plugins will there be?

laoneo:
Actually there are three, but the list can grow when extension devs do start to create new ones, there is not a fixed amount.

dneukirchen:
I think we can remove the "actual" preview column to get more space for editing. Nobody needs the image 2 times on the page.

dneukirchen:
I had this layout in mind: 
bildschirmfoto 2017-05-19 um 19 32 35

ciar4n:
Considering we go down to a single preview, is there any need for the preview title ('edit area', 'actual preview')? Personally I would say not as it is pretty self explanatory.

ciar4n:
media-edit1

dneukirchen:
Agree, no need for the title. 
Looks good. thank you @ciar4n  

@laoneo what do you think?

ciar4n:
I have applied this to the crop plugin however don't merge as needs some changes

laoneo:
For me it looks fine, cleane up and simple. I'm not sure if we can skip the editing area for the crop plugin. @dgt41 guess your input is needed here.

dneukirchen:
We dont remove the "editing" area, we remove the "preview" area. Do you think we need it?

laoneo:
Resize and rotate has only preview? Will that then become the editing area when they are activated?

dgrammatiko:
> We dont remove the "editing" area, we remove the "preview" area. Do you think we need it?

If we're gonna go the no action button, then the 2 images is kinda expected, so the user understands whites happening in the window especially for crop. So personally I would add back the action button for crop

Other than that, of course I like the new layout...

ciar4n:
Ready for testing

laoneo:
For me it is ok.

Just a minor not (as usual):
I would put the controls on the left side. When I change the plugins or do a toolbar action like save, then I'm on the left side, but when I want to change something in the controls, then I need to go to the right side of the screen. It gets somehow nervous.

dgrammatiko:
> I would put the controls on the left side. 

Right is the new left. What I mean is that in current (mobile) era the apps tend to have the most prominent actions on the right not on the left...

laoneo:
Sounds as an argument :+1: . I still think the stuff needs to closer to each other, the following gif shows what I'm talking about.
webinstaller-3
 But if you guys are comfortable with it, then I'm fine merging it.

dneukirchen:
I agree with allon. The mouseflow is not optimal in the current layout. I will merge it anyway and we will check the user-feedback from jab and ux-testing.

dneukirchen (approved):

dneukirchen:
thank you sooo much @ciar4n 👍

PR #208, opened 2017-05-16
Use only the language strings which are actually used

laoneo:
Cleans up the language file with strings only used by the new media manager.


dneukirchen (approved):
Thank you allon.

PR #211, opened 2017-05-17
Use checkbox toggle to select/unselect single or multiple items

dneukirchen:
This adds the ability to select/unselect an browser item using the checkbox on the item.


ciar4n:
Works well.

PR #219, opened 2017-05-18
The root information should be compiled in the local plugin

laoneo:
The local root path should be compiled in the plugin and not the model as the model doesn't know if the adapter is working on a local fileysystem or a cloud.


dneukirchen (approved):

dneukirchen:
thx @laoneo

Issue #230, opened 2017-05-25
Filenames with spaces cause 403


dgrammatiko:
### Steps to reproduce the issue

Try to upload a file that the filename contains a space

### Expected result

Uploads normally 

### Actual result

Error 403 on ok exits app!

dneukirchen:
@laoneo can you take a look at this? seems to be server related.

laoneo:
Somebody at JAB fixed it, so waiting for the pull request.

dneukirchen:
perfect, thx

ylahav:
It seems to me that the PR #261 and #262 can be accepted.
BTW - Can I run the tests for code convention (or any other test) before requesting a PR?

laoneo:
Yes, do a `composer install` in the Joomla root dir. Then do `libraries/vendor/bin/phpcs --report=full --extensions=php -p --standard=build/phpcs/Joomla .`

laoneo:
Closing as the pr's got merged.

PR #245, opened 2017-06-12
Remove spaces from upload filenames

ylahav:
Pull Request for Issue #230 .

### Summary of Changes

when an uploaded file has spaces we remove the space.

### Testing Instructions

try to upload several files with different special characters like spaces, dashes etc.   

### Expected result

Uploaded file and no 403 error

### DIFF
--- a/libraries/src/CMS/Helper/MediaHelper.php
+++ b/libraries/src/CMS/Helper/MediaHelper.php
@@ -152,6 +152,7 @@ class MediaHelper
 
 			return false;
 		}
+		$file['name'] = str_replace(' ', '', $file['name']);
 
 		jimport('joomla.filesystem.file');
 


laoneo:
Thank you very much for the pr. Unfortunately this function in the model should fix that as well https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/administrator/components/com_media/models/api.php#L193. But because the code in the controller got executed first it throws an error. So we need to move the getSafeName function back to the controller and execute it before https://github.com/joomla-projects/media-manager-improvement/blob/milestone1/administrator/components/com_media/controllers/api.json.php#L215.

kasvith:
Users may want to keep files and folders with spaces like in a typical filesystem IMO

laoneo:
There must be a reason why it is stripped out in core, this code was taken from the old media manager itself.

ylahav:
Kasvith - Spaces in file names are a nightmare with the web; you deal with %20 and other nonsense when spaces are in file names...

Allon - I think I understand your point. I will try to understand how the function getSafeName is used before making changes...  I think we have 2 questions:
1. Do we handle the 'spaces' issue in getSafeName, and (if I understood correctly),
2. Where the code of getSafeName should be... in models/api.php or somewhere in the controller. From your message I understand that the function was in the controller and moved to the model. If that the case - who made the move?

laoneo:
1. It should be handled where it is now, because the MediaHelper is probably used on other extensions, so we can't change the behavior. So we must provide a correct file name to that function, otherwise it fails.
2. I made that move, but can be moved back when required.

schnuti:
Be careful when changing anything around filenames. It has bin discussed a lot of times , including the spaces, ending with -> Do not change. There are always some BUTs in this OS ... and in this language .... restricting things. One simple ex. is ,as a user I have to be careful using MyImage.jpg and myimage.jpg. It can be the same image .. or not. In other examples some characters are allowed, some not.

dneukirchen:
I have to agree with @schnuti here... Names should not be modified until we have a very good reason to do so. 

The error is caused by the whitespace check in the mediahelper::canUpload method (Line 160..). 
Why do we need to replace whitespaces?

laoneo:
That's the logic from the old media manager. To stay on the safe side I thought it would be best not to change that logic as there are reasons why it is that way. If these reasons are still accurate is another question. I just want to be safe here as we had a vulnerability in 3.7.1 in the media manager. If somebody wants to spend time to carefully review the upload logic, then I'm fine. But till then I suggest to stick to the logic of the old media manager.

schnuti:
@dneukirchen 
Then you misunderstood me. I say - Do not change anything around the filename checks and handling without a very good reason! Try to upload a file with blanks and some umlaut with the old MM. - is added for the whitespace and umlaut is **removed**. I'm sure there are cases where you'll have problems with a blank in the filename.
Another problem is that character encodings are different in different file systems. I hope the Media Manager soon come to use DB tables. Then you'll have file names utf8 encoded in a table. And now?

I use tables to manage my images and use the item-alias method for the stored images file names. So far this works fine for my own Western Latin characters. I know it also works for cyrillic characters. All translates to a-z,0-9 and -. I have no clue about asian languages.

dneukirchen:
> If somebody wants to spend time to carefully review the upload logic, then I'm fine. But till then I suggest to stick to the logic of the old media manager.

I prefer to not modify the names by default, but im ok with it.

So how do we proceed here @laoneo @ylahav? Quickest solution is to just str_replace in controller. Im fine with that.

kasvith:
I vote for same, i would not like to rename files while uploading

laoneo:
I suggest this https://github.com/joomla-projects/media-manager-improvement/pull/245#issuecomment-308125027

ylahav:
I will make the change in the controller... BUT let see that I understood correctly :-)
We remove the space according @laoneo comment

dneukirchen:
Yes, @ylahav please move getSafeName from model to the controller. Is that ok for you?
getSafeName() is basicallay an input filter similar to the filters of JInput, so its ok for me if we do that in controller.

ylahav:
BTW the best solution here is to store the 'filename' as is in DB and create a filename and refer it to the orig 'filename'.
This is more complex implementation because DB is involved and more than local changing....
and, yes @dneukirchen I will do it during the coming week

laoneo:
Thank you.

schnuti:
Not change file names?
Change a jpg's filename to Übergrößenträger.jpg and upload it with the old MM. Got it?

Are there DB tables involved  in the new MM at this stage? If so then @ylahav has the solution.

You should also discuss if some other normalizing of filenames is useful? e.g. always jpg and png and not JPG, Jpeg, jpeg, JPEG, Jpeg, PNG, Png or whatever is allowed on the IS level. It looks more professional to me.

dneukirchen:
> Change a jpg's filename to Übergrößenträger.jpg and upload it with the old MM. Got it?
-- you

> Names should not be modified until we have a very good reason to do so.
-- me

Umlauts are a good reason. So we are on the same page here... 

> Are there DB tables involved in the new MM at this stage? If so then @ylahav has the solution.

No,DB tables are currently not involved. We had to postpone some planned features in favor of a realistic timeline. I think we all agree, that we would not have this discussion with a db layer. 

Thank you for your input! 
It would be nice if you create an issue for your normalization suggestion. 👍

schnuti:
@dneukirchen 
Do you not want to understand? I still say - Do not allow **other characters** in filenames without very good reasons.
i.e Übergrößenträger.jpg should be translated to uebergroessentraeger.jpg as those characters are allowed in the old MM, every known OS, links, every known communication channel aso. Blanks (whitespace)  is Not allowed everywhere.
What characters possible to use, is changing rapidly and in the end we'll use utf-8 encoding everywhere. A "conservative" system has though to consider older systems in use. Ask Microsoft when Windows and Unix character encoding will be interchangable.

schnuti:
@dneukirchen I forgot to ask. Did you test the Übergrößenträger. I always found that result as a very big bug. It's time to resolve it.

ylahav:
@schnuti - I will test it after making the change....

kasvith:
also add support to unicode texts like සිංහල භාශාව

laoneo:
@kasvith please open another issue, would not like to mix things here. One issue one problem. Thanks for understanding.

kasvith:
i hope this is also regarding the issue, 
if we convert names like 

> Übergrößenträger.jpg should be translated to uebergroessentraeger.jpg 

It should also support other languages too

ylahav:
@laoneoI, two questions:
1. look in the 'getSafeName' function and saw that it is called only from the 'api.php', so why to moved it to the controller?
2. And, how this change will be aligned with #260 ?

laoneo:
1. Because I expect that it will strip out the white spaces in the same way as in the old media manager.
2. At the moment we don't care in this pr about #260. One pr per problem :)

laoneo:
Closing this one as it get fixed by #261 and #262.

PR #248, opened 2017-06-13
Check if the mime_content_type exists

laoneo:
Pull Request for comment in main repo https://github.com/joomla/joomla-cms/issues/16586#issuecomment-307553504.

### Summary of Changes
Adds a check if the function `mime_content_type` exists.


brianteeman:
this will only partially address the issue. as this is now a requirement for joomla we need to add it to the preinstallation checks, system info reports etc etc

laoneo:
It is not a requirement, more of a nice to have.

brianteeman:
it is a requirement as its completely broken without it

laoneo:
What is broken without it? Just the mime type is not displayed.

zero-24:
What about this method as it has some more fallbacks: https://github.com/joomla/joomla-cms/blob/staging/libraries/cms/helper/media.php#L59 ?

Need to be declared as public first but can be done.

laoneo:
@zero-24 good catch, will change it to that function then.

laoneo:
Updated to the mediahelper function

dneukirchen:
Thx @laoneo

dneukirchen (approved):

PR #253, opened 2017-06-14
Feature/multi adapter

dneukirchen:
Simple implemtation of multi adapter support in client.
I dropped my first idea to have nested vuex modules for each disk, but i will keep the changes in the `multi-adapter-support` branch. Perhaps we have to switch to that method later

Please note: because of the path changes the image preview and some file operations do not work atm.

### DIFF
--- a/administrator/components/com_media/resources/scripts/app/Api.js
+++ b/administrator/components/com_media/resources/scripts/app/Api.js
@@ -130,6 +130,10 @@ class Api {
 
         item.directory = path.dirname(item.path);
 
+        if(item.directory.indexOf(':', item.directory.length - 1) !== -1) {
+            item.directory += '/';
+        }
+
         return item;
     }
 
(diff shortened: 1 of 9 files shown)


laoneo:
The images are not displayed in the list.

Additionally the tree looks a bit strange with this line crossing the icons
image

laoneo:
The database install/update script needs to be updated, setting the title information, because when the local plugin doesn't have a title saved in the params, then it will not displayed in the tree.

dneukirchen:
> The images are not displayed in the list.

Yes. This is because of the path changes. I could have removed it in the client, but for remote providers it will not work (when the resources are not public). So we have to make some decisions here before i will implement it.

> Additionally the tree looks a bit strange with this line crossing the icons

Jups. WIP. => we need styling before merging gsoc to ms1

> The database install/update script needs to be updated, setting the title information, because when the local plugin doesn't have a title saved in the params, then it will not displayed in the tree.

Yes, but imo this is not related to this PR as it focusses on the client. I already told @kasvith that the defaults are missing.

laoneo:
Merged it so we can move on. Thanks!

kasvith:
@dneukirchen it has default values in local.xml

dneukirchen:
Yes, but 

> The database install/update script needs to be updated, setting the title information, because when the local plugin doesn't have a title saved in the params, then it will not displayed in the tree.

PR #262, opened 2017-06-24
Update Model/Api.php Controller/Api.php

ylahav:
Pull Request for Issue #230 .

Summary of Changes

when an uploaded file has spaces it is replace by a underscore (same way for folder creation).
It was done by:

moving 'getSafeName' function from the model Api.php to the controller Api.ph
Call 'getSafeName' in the controller API in two places: (a) creation of a file or directory, and (b) uploading a new image
remove the call of 'getSafeName' in the model
Testing Instructions

try to upload several files with different special characters like spaces, dashes etc.

Expected result

Uploaded file and no 403 error
Actual result

I made some tests in "old Joomla (3.7.2), before my changes and after it and swa the following:

filename with english latters and one or more spaces
"old": an error - "Unable to create folder. Folder name must only contain alphanumeric characters and no spaces."
J4 before the change - 403
J4 after the change - all spaces are replaced with '_'
file name with special charcter (e.g. Hebrew)
"old": an error - "Unable to create folder. Folder name must only contain alphanumeric characters and no spaces."
J4 before the change - nothing happens
J4 after the change - nothing happens
directory name with spaces
"old": an error - "Unable to create folder. Folder name must only contain alphanumeric characters and no spaces."
J4 before the change - all spaces are replaced with '_'
J4 after the change - all spaces are replaced with '_'
directory with special charcter (e.g. Hebrew)
"old": an error - "Unable to create folder. Folder name must only contain alphanumeric characters and no spaces."
J4 before the change - nothing happens
J4 after the change - nothing happens

### DIFF
--- a/administrator/components/com_media/Model/Api.php
+++ b/administrator/components/com_media/Model/Api.php
@@ -120,8 +120,6 @@ class Api extends Model
 	 */
 	public function createFolder($name, $path)
 	{
-		$name = $this->getSafeName($name);
-
 		$this->adapter->createFolder($name, $path);
 
 		return $name;
@@ -143,8 +141,6 @@ class Api extends Model
 	 */
 	public function createFile($name, $path, $data)
 	{
-		$name = $this->getSafeName($name);
-
 		$this->adapter->createFile($name, $path, $data);
 
 		return $name;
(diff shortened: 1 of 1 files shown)


laoneo:
Thank you very much!

PR #263, opened 2017-06-29
File serve from multi adapter 

kasvith:
### Summary of Changes
- Added file serving method to API
- Added file serving invoking to Controller

### Testing Instructions
- Log in to backend
- Navigate to following URL 

http://localhost/media-manager-improvement/administrator/index.php?option=com_media&format=json&url=true&task=api.files&path=local:/banners/banner.jpg

### Expected result
- AJAX response with data containing a url to banner.jpg


kasvith:
Tarvis will fail as i ve changed ctr a bit
Once code is acceptable i will fix them and we can  merge then

laoneo:
I'v merged the milestone1 branch into the gsoc17. Please merge now the gsoc17 branch into yours. Then the commit log should show up fine. Hopefully.

kasvith:
@laoneo done, can you check it?

kasvith:
tarvis will fail as i have not updated test cases

.gitignore
```diff
@@ -209,6 +209,7 @@ Desktop.ini
 /libraries/vendor/zendframework/zend-diactoros/CONTRIBUTING.md
 /libraries/vendor/zendframework/zend-diactoros/mkdocs.yml
 /libraries/vendor/zendframework/zend-diactoros/README.md
+/plugins/filesystem/dropbox/composer.lock
```
laoneo: Please remove that line.

administrator/components/com_media/Adapter/AdapterInterface.php
```diff
@@ -149,4 +149,17 @@ public function move($sourcePath, $destinationPath, $force = false);
 	 * @throws  \Exception
 	 */
 	public function copy($sourcePath, $destinationPath, $force = false);
+
```
laoneo, line 152: One empty line to much.

administrator/components/com_media/Adapter/AdapterInterface.php
```diff
@@ -149,4 +149,17 @@ public function move($sourcePath, $destinationPath, $force = false);
...
+	 * Returns a permanent link for media file.
+	 *
+	 * @param   string  $path  The path to file
+	 *
+	 * @return string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws FileNotFoundException
+	 */
+	public function getPermalink($path);
```
laoneo: As discussed I would change that to getUrl or something more meaningful.

administrator/components/com_media/Controller/Api.php
```diff
@@ -151,7 +165,7 @@ public function files()
 			// Return the data
 			$this->sendResponse($data);
 		}
-		catch (MediaFileAdapterFilenotfoundexception $e)
+		catch (FileNotFoundException $e)
```
laoneo: Can you make a pr with this fix against the milestone1 branch? We need to have it fixed there as well.

administrator/components/com_media/Controller/Api.php
```diff
@@ -104,7 +105,20 @@ public function files()
 			switch ($method)
 			{
 				case 'get':
-					$data = $this->getModel()->getFiles($adapter, $path, $this->input->getWord('filter'));
+					$option = $this->input->getWord('opt');
```
laoneo: I would rather make it as an option that the get response should include the permalink or not in the returned data. IMO the GET response should always return the same data, but with some more details or not.

plugins/filesystem/local/Adapter/LocalAdapter.php
```diff
@@ -544,4 +554,19 @@ private function moveFolder($sourcePath, $destinationPath, $force = false)
...
+	 *
+	 * @param   string  $path  Path of the file relative to adapter
+	 *
+	 * @return string
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	public function getPermalink($path)
+	{
+		return \Joomla\CMS\Uri\Uri::root() . \JPath::clean($this->filePath . $path);
```
laoneo: No need to add here the full namespace as you made the use statement already.

laoneo (changes requested):

administrator/components/com_media/Controller/Api.php
kasvith: Sure

administrator/components/com_media/Controller/Api.php
kasvith: I did not get that, anyway do we want an assoc array?

administrator/components/com_media/Adapter/AdapterInterface.php
kasvith: it is updated removed line follows that

administrator/components/com_media/Controller/Api.php
kasvith: No actually as we discussed we only need to get URL iff requested, so IMO we should not send it along file attributes at all, only when it requested we send URL along with request for selected file

kasvith: If you want we can create another task

kasvith: @dneukirchen do we need something like url => "path/to/url" or just "path/to/url" along with data in AJAX response ?

administrator/components/com_media/Controller/Api.php
```diff
@@ -104,7 +105,20 @@ public function files()
...
+						case 'permalink':
+							$data = $this->getModel()->getUrl($adapter, $path);
+							break;
+
+						default:
+							$data = $this->getModel()->getFiles($adapter, $path, $this->input->getWord('filter'));
+							break;
+
+					}
+
```
dgrammatiko: @kasvith I think an if else is better here (`if (isset($option) && $option === 'permalink')`

administrator/components/com_media/Controller/Api.php
kasvith: i ve made pr

administrator/components/com_media/Controller/Api.php
kasvith: ya an if would do the job also

administrator/components/com_media/Controller/Api.php
laoneo: For example the google calendar api has a flag `alwaysIncludeEmail` which returns the email addresses in the response or not https://developers.google.com/google-apps/calendar/v3/reference/events/list. I would rather use something like that then returning a totally different response on a GET request for the same endpoint.

administrator/components/com_media/resources/scripts/store/state.js
```diff
@@ -23,6 +23,4 @@ export default {
     selectedDirectory: options.providers[0].name + ':/',
     // The currently selected items
     selectedItems: [],
-    // The currently selected items
-    selectedItems: [],
```
laoneo: Why is this file modified?

administrator/components/com_media/Adapter/AdapterInterface.php
```diff
@@ -149,4 +149,16 @@ public function move($sourcePath, $destinationPath, $force = false);
 	 * @throws  \Exception
 	 */
 	public function copy($sourcePath, $destinationPath, $force = false);
+
+	/**
+	 * Returns a permanent link for media file.
```
laoneo: Please add here a better comment what's the purpose of this function.

administrator/components/com_media/Controller/Api.php
```diff
@@ -104,7 +105,11 @@ public function files()
 			switch ($method)
 			{
 				case 'get':
-					$data = $this->getModel()->getFiles($adapter, $path, $this->input->getWord('filter'));
+					$options = array();
+					$options[] = $this->input->getWord('opt');
```
laoneo: I would make it an indexed array, where the key is here url and the value true or false.

administrator/components/com_media/Controller/Api.php
laoneo: Please update your branch with the gsoc17, merged your pr already.

administrator/components/com_media/Model/Api.php
```diff
@@ -120,23 +121,30 @@ public function getFile($adapter, $path = '/')
...
+		// Add adapter prefix to all the files to be returned
 		$files = $this->getAdapter($adapter)->getFiles($path, $filter);
 
 		foreach ($files as $file)
 		{
 			$file->path = $adapter . ":" . $file->path;
 		}
 
+		// Add url info to a file when option is set
+		if (in_array("url", $options) && count($files) === 1 && $files[0]->type == "file")
```
laoneo: ` && count($files) === 1 && $files[0]->type == "file"` can be removed, if somebody wants to have it for all files, then it is in his responsibility. We should not force that. I'm aware that it can be a performance issue, but so it is.

laoneo (changes requested):

administrator/components/com_media/Controller/Api.php
kasvith: ahh thats a good idea

administrator/components/com_media/resources/scripts/store/state.js
kasvith: tried some experiment to fix create folder model, will revert it

administrator/components/com_media/Model/Api.php
kasvith: IMO method getUrl only should act on files, in here we are actually giving a folder path as the input path to getUrl and 3dp may be mislead to implement it for a whole folder.

If someone break this plugins will be malfunction IMO

administrator/components/com_media/resources/scripts/store/state.js
kasvith: there was a duplicated variable i removed it

administrator/components/com_media/Model/Api.php
kasvith: even we added url for files in a directory if it contain a directory it makes no sense, IMO we should force them to apply this only on a single file. Since 3dps will never touch this we can make em force to avoid unnecessary checks in api

administrator/components/com_media/resources/scripts/store/state.js
laoneo: I would do that in another pr against the milestone 1 branch. One thing in one pr.

tests/unit/suites/plugins/filesystem/local/LocalAdapterTest.php
```diff
@@ -26,6 +26,13 @@ class LocalAdapterTest extends TestCaseDatabase
 	private $root = null;
 
 	/**
+	 * The image folder path related to root
+	 * 
+	 * @var null
+	 */
+	private $image_path = null;
```
laoneo: Please camel case

laoneo (changes requested):

laoneo:
Would be good to also add a test for getUrl

administrator/components/com_media/resources/scripts/store/state.js
kasvith: Ahh sorry about that

administrator/components/com_media/Model/Api.php
```diff
@@ -120,23 +121,32 @@ public function getFile($adapter, $path = '/')
...
 		foreach ($files as $file)
 		{
+			// If requested add options
+			if (isset($options['url']) && $options['url'] == 'true' && $file->type == 'file')
+			{
+				$file->url = $this->getUrl($adapter, $file->path);
+			}
+
 			$file->path = $adapter . ":" . $file->path;
+
```
laoneo: Empty line not needed

administrator/components/com_media/Model/Api.php
```diff
@@ -120,23 +121,32 @@ public function getFile($adapter, $path = '/')
...
 		}
 
+
```
laoneo: Empty line not needed

laoneo (changes requested):

laoneo:
What did you do for changes in mediamanager.js?

kasvith:
@laoneo i have run build-js again after changing the state.js

laoneo:
But you reverted that change, so you have to revert that file then as well.

administrator/components/com_media/Model/Api.php
```diff
@@ -120,20 +121,27 @@ public function getFile($adapter, $path = '/')
...
+	public function getFiles($adapter, $path = '/', $filter = '', $options = array())
 	{
-		// Add adapter info to all the files to be returned
+		// Add adapter prefix to all the files to be returned
 		$files = $this->getAdapter($adapter)->getFiles($path, $filter);
 
 		foreach ($files as $file)
 		{
+			// If requested add options
+			if (isset($options['url']) && $options['url'] == 'true' && $file->type == 'file')
```
laoneo: I would not test if the string is `'true'`. I would just do `if (isset($options['url']) && $options['url'] && $file->type == 'file')`.

administrator/components/com_media/Controller/Api.php
```diff
@@ -105,7 +105,12 @@ public function files()
 			switch ($method)
 			{
 				case 'get':
-					$data = $this->getModel()->getFiles($adapter, $path, $this->input->getWord('filter'));
+					// Grab options
+					$options = array();
+					$options['url'] = $this->input->getWord('url', '');
```
laoneo: I would cast that here to boolean, see my comment below why.

laoneo (changes requested):

laoneo:
Thank you very much!

kasvith:
thanks @laoneo for spending time

dneukirchen:
gj @kasvith

PR #266, opened 2017-07-07
Thumbnails support

kasvith:
### Summary of Changes
After multi adapter architecture was introduced Thumbnail displaying was broken.
This PR will bring it back to the system
It was used to have original images as the thumbnail before. 

This PR includes a **thumb_path** field which can be used to setup an actual thumbnail instead of the image itself.

**At the moment** this only support images, but it can be configured to support any file or folder so 3rd party developers can provide nice themes for their adapter.

**At the moment the PR uses actual image instead of thumbnail since The thumbnail system not implemented yet for Local File System**


administrator/components/com_media/package.json
```diff
@@ -10,7 +10,9 @@
     "build-js": "browserify ./resources/scripts/mediamanager.js > ../../../media/com_media/js/mediamanager.js"
   },
   "dependencies": {
+    "amdefine": "^1.0.1",
     "babel-polyfill": "^6.23.0",
+    "has-flag": "^2.0.0",
```
dneukirchen: Are those 2 new deps used anywhere?

administrator/components/com_media/package.json
```diff
@@ -10,7 +10,9 @@
     "build-js": "browserify ./resources/scripts/mediamanager.js > ../../../media/com_media/js/mediamanager.js"
   },
   "dependencies": {
+    "amdefine": "^1.0.1",
```
dneukirchen: Where is this used?

administrator/components/com_media/package.json
kasvith: i could not do npm run build without those, i told you before about this issue and also @dgt41

administrator/components/com_media/package.json
kasvith: i could not do npm run build without those, i told you before about this issue

kasvith:
There was a problem in my setup it seems, now it is fixed, removed deps

dneukirchen:
Great, thx.

dneukirchen (approved):

dneukirchen:
thank you @kasvith

PR #267, opened 2017-07-09
Foo adapter removal

kasvith:
### Summary of Changes
We don't need Foo adapter since a **dropbox adapter** on the way


laoneo:
I would remove it when we have a working cloud plugin, not before.

PR #271, opened 2017-07-11
Feature/dropbox plugin

kasvith:
Pull Request for Issue # .

### Summary of Changes
- Removed name assign after createFile and createFolder. They actually does not return any info, they are void methods. This issue will prevent getting uploaded file info as the **$name** is set to null by those lines


kasvith:
Opps

PR #273, opened 2017-07-11
removed name variable from controller when create a file or folder

kasvith:
Pull Request for Issue # .

### Summary of Changes
In Controller there was setting name after createFile or createFolder.
This is not required as we do not return anything from those methods.
Need to remove them to prevent making name null


dneukirchen (approved):

laoneo:
Should this pr not be against milestone1 branch?

kasvith:
it has that issue too, yes it is against it

kasvith:
i will do a pr against it wait

PR #275, opened 2017-07-17
Fix/ Api returns null on names

kasvith:
Pull Request for Issue # .

### Summary of Changes
In Controller there was setting name after createFile or createFolder.
This is not required as we do not return anything from those methods.
Need to remove them to prevent making name null


laoneo:
Thanks

kasvith:
@laoneo can you merge this to gsoc17 branch too

PR #276, opened 2017-07-19
Feature/dropbox plugin

kasvith:
### Summary of Changes
- Removed **Foo adapter** from Joomla!
- Added a Mock Dropbox plugin for *testing* with a simple cache
**Please do not modify files in vendor folder since i made some changes to methods there which is required**

### Testing Instructions
First please follow the procedure
- Go to https://www.dropbox.com/developers/apps/create
- Choose **Dropbox API** and select **Full Dropbox**, then give a name and create app.
- Then you will redirected to another page
- Click **Generate** under **Access Token**
- Copy the generated token

- Go to Joomla!
- Enable *Dropbox* plugin
- Go to plugin settings
- Under *Dropbox API Key* paste the token you got
- Save
- Go to Media Manager
- Click on *Dropbox*
- Wait

### Expected result
- You may see your files from dropbox
- You can also create folders, upload files to Dropbox

### Actual result
- It should work

- Plugin is purely for testing purpose


kasvith:
**Take Note : Drone always fail because the vendors have not followed our coding standards, since i ve modified some files in vendors DO NOT ATTEMPT A COMPOSER INSTALL EITHER**

laoneo:
Can you make a new pr with only the thumbnail feature?

kasvith:
@laoneo these thumbnails are generated by the dropbox itself, I only saved them, do you like to have such a feature for Local File System ?

laoneo:
On a first glance it looked like that you have implemented thumbnails. What are they for then in the dropbox adapter? Did you change the client to use thumbnails? I suggest, either way we introduce thumbnail support in the client or not, but having it here without real usage is for me confusing.

kasvith:
Dropbox provide it's own method to get the thumbnails via API, but at the moment they only serve it as a stream. So the developer must take the stream and dump it to a file to use it in the program.

In future they are supporting to provide a thumbnail url for the API. But at the moment they don't.
So i developed a simple thumbnail cache thats only for the dropbox adapter.

In future for even Joomla Media Manager we only need to provide a cache for the Local FS which will be shipped with the joomla, other 3dp developers may need to get the thumbnail in a separate way, maybe through a custom API.

kasvith:
I only developed this for reduce the load time of images, unless it take some time to load the preview of the files in dropbox (usual delay in the typical dropbox website is also applied here)

We must keep in mind

**Different adapters have different methods to obtain thumbnails, some of them are already generated for many sizes(in here I have used a concrete size). So we should only focus on Local FS because it is shipped with Joomla. It make no sense if your provider gives you a direct thumbnail link and we have a cache for all thumbnails in joomla**

kasvith:
Simply we cant have a common cache for all the adapters. They behave differently, but we need a pretty good cache for the **Local FS**

laoneo:
I guess this needs to be discussed separately how we want to support thumbnails.

kasvith:
Yes for the *Local FS* we need to do a discussion, but for 3dp plugins they usually provide direct links to the thumbnails(Dropbox says they will provide that facility asap) so we don't even need to store like this(it is temporary :) )

laoneo:
Then it would be good to add some comments to the code that the thumbnails are just for demo purposes and are not yet used in the client.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
```diff
@@ -0,0 +1,388 @@
...
+	 */
+	private $dropbox = null;
+
+	/**
+	 * DropboxAdapter constructor.
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct($apiToken)
+	{
+		$this->client  = new \Srmklive\Dropbox\Client\DropboxClient($apiToken);
```
laoneo: Move them to their own function like getClient and call it only when needed. Plugins can be instantiated without even being used.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
```diff
@@ -0,0 +1,388 @@
...
+	 *
+	 * @return string
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function getThumbnailUrl($id, $timeModified , $path)
+	{
+
+		$name = explode(":", $id)[1];
+		$timeStamp = strtotime($timeModified);
+		$filePath = \JPath::clean(JPATH_PLUGINS . '/filesystem/dropbox/.thumb_cache/' . $name . $timeStamp . '.jpg' , '/');
```
laoneo: Path should be /media/plg_filesystem_dropbox/thumbs and not in the plugins folder.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
```diff
@@ -0,0 +1,388 @@
...
+	 * @param   string $path The path to the file or folder
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  \Exception
+	 */
+	public function delete($path)
+	{
+		$response =	$this->client->delete($path);
```
laoneo: Too many spaces

plugins/filesystem/dropbox/dropbox.php
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @since    __DEPLOY_VERSION__
+	 */
+	public function onFileSystemGetAdapters()
+	{
+		$apiToken = $this->params->get('api_token');
+
+		if (!$apiToken || strlen($apiToken) <= 0)
+		{
+			throw new \Exception("API token was not found");
```
laoneo: Don't think so that we need to throw here an exception. Just return an empty array.

laoneo (changes requested):

plugins/filesystem/dropbox/dropbox.php
kasvith: We need to show a warning for the user so he can directly fix that

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
kasvith: So it means we need to do it in the install script to make this folder ?

laoneo: Just on the fly.

plugins/filesystem/dropbox/dropbox.php
laoneo: Then make that field mandatory.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
kasvith: my ide(PHPStrom) shows only 1 space

laoneo: It's a tab

plugins/filesystem/dropbox/dropbox.php
kasvith: can you give me an example ? how to make it mandatory ?

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
kasvith: should they be public ?

plugins/filesystem/dropbox/dropbox.php
laoneo: Add the required flag to the field description in the manifest XML file.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
laoneo: Nope, private.

plugins/filesystem/dropbox/Adapter/JoomlaDropboxAdapter.php
kasvith: you are suggesting me to create a folder in media/plg_filesystem_dropbox ? and refer that as the cache right?

Since this was a pure demo i thought to keep all the files within plugin folder, it will be easy to remove it when its not needed

(We don't ship this plugin I assume)

laoneo: Despite it's only for demo/testing purposes, we should follow the Joomla standard.

laoneo:
I tested it, but when the dropbox plugin is enabled, then the local one will not be displayed anymore...if you can fix that, then it is ready to merge.

kasvith:
In my setup there was no such a case, could you elaborate it more

laoneo:
On the tree bar on the left side, I see only the dropbox folders, but not anymore the ones from the local file system

laoneo:
Issue was sorted out. I'm going to merge this one that we can head on. Before the merge into the core, we need to remove it again as we probably don't want to ship a dropbox plugin in core.

laoneo:
Thanks. Really happy to have a working cloud plugin.

kasvith:
Thanks @laoneo

masteranalyze:
Great work,finally joomla 4 at least will have media manager that will support basiccaly from core the cloud file sync.

As they are alot of Cloud providers,i think is good to be implemented plugins for most used like : google drive,dropbox,microsoft one drive,etc.

In order to practically integrate all cloud providers on the go and also having the files encrypted on  your cloud provider not only synced,their is this very good thing called : https://nextcloud.com/ ,and they add more and more providers,and nextcloud operates with secured encrypted cloud providers,that means if you put an file with you and your child,you will also have privacy,as google,dropbox,microsoft won`t have your key for decrypting it.

The major problem till know,it was that nobody thinked on an system or plugin for media manager for joomla,to work with your nextcloud.

Nextcloud also supports : ftp,Webdav,localfiles,etc and they are integrating on the go all cloud providers,so instead of having 100 plugins for the media manager,if somebody can figure it out somekind of cloud sync via webdav or something,between joomla and nextcloud,to easily share your files from nextcloud into joomla it will be great stuff.

The thing is ,can somebody develop some kind of plugin with the new media manager for next cloud?

That way,you can add your files in next cloud secured and encrypted via ftp,cloud provider,aws,dropbox,google,etc and then share it into Joomla.

Right now basically they are extension for joomla in witch you can put and share the link from nextcloud,remote into the extension for example in docman that is possible.

The problem is,that if you want to put 1000 images and share them or even 100 you have to put manually link by link,is an big problem,because you have to put them 1 by 1,and that is insane in this century of speed.

But for example if you have the files synced like it was in there into the media manager,all the files,can quickly be added,and you don`t have to put them 1 by 1,you do this in batch.

Nextcloud just like joomla is fully open source,and updated constantly and is most secure,is like having your own dropbox secure service,it provides privacy to your files,except the sync between cloud providers.

I mean will be much more simpler,then having to build plugins for all cloud providers,just an plugin for nextcloud to work with joomla,and the problem is solved with all cloud providers and more+ (ftp,webdav,etc)

kasvith:
@masteranalyze thanks for the feedback.
Yes if they use OAuth2.0 and if they can follow our adapter architecture anyone can create a plugin to connect it to Media Manager

masteranalyze:
great work done kasvith,joomla deserves an new media manager adapted for today needs and future.

Can you check out this,i m not sure,but i think they have this oauth2.0:

https://help.nextcloud.com/t/nc12-build-in-oauth2-client-authentification/14852

So in this case,it can be implemented?

the nice thing about nextcloud is you have already all cloud providers(so no need to write plugin for each cloud provider),you have webdav,ftp,etc you have ability to store securely and encrypted your files into the cloud providers(as privacy with Acta and all those that wants to invade our private life without our consent,is an very important thing,so you have privacy on all your files) .

in wp,there is no nextcloud integration,witch can also can boost new users adoption of joomla.

Basically if this is possible and working,with one plugin just for nextcloud,the new joomla 4 media manager can be an awesome feature integrating most cloud providers,providing security and privacy,and not only cloud providers,but users can connect ftp,sftp,wevdav and more into nextcloud.

let me know what you think,after you research it.

laoneo:
Just to be clear here, the core will not ship with any cloud plugin. We will only ship the local plugin which gives access to the local file system. The dropbox plugin acts for the time being as proof that the whole architecture works. On some point we will move it to its own repo. This project should encourage developers to build plugins for any cloud system they want/need.

Just a side not to your nextcloud request. It should be possible to write a cloud plugin for your it, but it will not be distributed as part of the core.

masteranalyze:
that is what matters,at least to be possible,it does not matter who writes that plug-in,the point is to be possible at least then to work.

this plugins as they will be made by 3rd and not ship into the core,the devs will be able to submit this plugins for the new media manager in joomla extension directory or?

i mean i understand you want to make the core as simple as possible,as shipped with different plugins can require more testing and quality assurance of the code,but it should be an place where devs of this plugins to submit it,just like in the extension directory of joomla or something similar.

i will write on nc and open an new issue and inform them about this new possibility from the joomla 4 new media manager,but can you point me an link or some sort of code docunentation,that should be followed by devs in order to create this?

kasvith:
Hi @masteranalyze thank you for the feedback.

As @laoneo mentioned we ship Joomla! without these plugins. We only ship Local Filesystem Plugin with Joomla 4.

You can simply create a Filesystem plugin using the documentation [here](https://docs.joomla.org/Cloud_File_Systems_for_Media_Manager) . After you creating it simply activate the plugin and it will be automatically integrate with the Media manager without extra steps. 

All you have to do is create the proper plugin.

For an example you can use this #286 PR to look how authorization works and so on. We will move this to a separate repo soon

laoneo:
I'm going to lock this pr. If you want to continue the discussion, please open a new issue. Thanks for your interest!

PR #280, opened 2017-08-16
OAuthCallback Handler for Media Manager

kasvith:
### Summary of Changes
Added OAuthCallback endpoint to Joomla!

From now on 3rd Party Plugin developers can simply can get data from a redirect without using any controller written by them.

Developers can set redirect uri for 
**_[site-name]/administrator/index.php?option=com_media&task=plugin.oauthcallback&plugin=[plugin-name]_**

To send a request to Joomla, you must also set the state parameter of your cloud provider.
This is used common across  cloud providers to keep the consistent of the data. In this case put *token* request by `\JSession::getFormToken()` as **state**  in order to access Joomla from this callback.

Once successfully CSRF token is verified, your plugin will receive all the data passed to the callback.
For that your plugin must implement the following method

```
public function onFilesystemOAuthCallback(\Joomla\Component\Media\Administrator\Event\OAuthCallbackEvent $event)
{
     	// your code
	
	$result = $event->getArguments('result', []); // prevent overwriting existing results if any
	$result[] = 'your result';
	$event->setArguments('result', $result);
}
```

**result** is a special variable you need to set. Do not use legacy return along with this. It will not work. In any case read this [gist](https://gist.github.com/kasvith/1217c31414cfcece82fd5565b70ff80a)

Result defines the way Joomla behaves after a successful method call.

- 'action'
   Set this to 
   - close : Closes a window which is opened by a Java Script using *window.open()*
   - redirect : Redirects to the page specified by the *redirect_uri*
   - control-panel : Redirect to control panel
   - media-manager : Redirect to Media Manager

- 'redirect_uri'
   Used with *redirect* action (required)

- 'message'
   Message need to be displayed after an action

- 'message_type'
   - warning
   - notice
   - error
   - message(or set empty)

### Testing Instructions
Please comment out CSRF checking in Controller/Plugin.php and start navigating callback using url structure provided above.


administrator/components/com_media/Controller/Plugin.php
```diff
@@ -0,0 +1,187 @@
...
+ */
+
+namespace Joomla\Component\Media\Administrator\Controller;
+
+defined('_JEXEC') or die;
+
+use Joomla\CMS\Controller\Controller;
+use Joomla\CMS\Plugin\PluginHelper;
+use Joomla\Component\Media\Administrator\Event\OAuthCallbackEvent;
+use League\Flysystem\Exception;
```
laoneo: Please use core exceptions.

administrator/components/com_media/Controller/Plugin.php
```diff
@@ -0,0 +1,187 @@
...
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
+	public function oauthcallback()
+	{
+		// Get the input
+		$input = $this->input->request->getArray();
+
+		// Check if CSRF was set in status or reject
+		if (!isset($input['status']))
```
laoneo: Is this not google drive specific?

administrator/components/com_media/Controller/Plugin.php
```diff
@@ -0,0 +1,187 @@
...
+		{
+			jexit("CSRF token was not found");
+		}
+
+		// Check if token match with the session
+		if (\JSession::getFormToken() != $input['status'])
```
laoneo: Is this not google drive specific?

administrator/components/com_media/Controller/Plugin.php
```diff
@@ -0,0 +1,187 @@
...
+
+			// Only import our required plugin, not entire group
+			PluginHelper::importPlugin('filesystem', $pluginName);
+
+			// Event parameters
+			$eventParameters = ['context' => $pluginName, 'parameters' => $input];
+			$event = new OAuthCallbackEvent('onFilesystemOAuthCallback', $eventParameters);
+
+			// Get results from event
+			$eventResults = (array) \JFactory::getApplication()->triggerEvent('onFilesystemOAuthCallback', $event);
```
laoneo: Please use the namespaced Factory.

administrator/components/com_media/tmpl/plugin/default.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
laoneo, line 1: When is this view used?

laoneo (changes requested):

administrator/components/com_media/Controller/Plugin.php
kasvith: nope

administrator/components/com_media/Controller/Plugin.php
kasvith: nope

administrator/components/com_media/tmpl/plugin/default.php
kasvith: when in action *close* used

administrator/components/com_media/Controller/Plugin.php
kasvith: Sorry it was done by IDE

administrator/components/com_media/tmpl/plugin/default.php
kasvith: it can be used by 3dps to close any new window opened by themselves with a single line of code

administrator/components/com_media/Controller/Plugin.php
laoneo: So what happens when a cloud doesn't support the status flag? I guess it will not work.

laoneo:
Thanks!!

kasvith:
You are welcome

dneukirchen:
thx @kasvith

PR #281, opened 2017-08-22
Multiple Account support for Media Manager

kasvith:
### Summary of Changes
- Added multiple accounts for File Adapters
- Updated Local/Dropbox plugin with the new improvement
- Updated file path returned by adapters to be in `[provider-name]-[adapter]:/path/to/file` format

This can be used to display rooted providers like

> - Local Filesystem
>   - images
>   - my images
> - Dropbox 
> - Google Drive

Now when obtaining **provider list** in the front end, instead of just Plugin names, all accounts that are associated with relevant provider will be listed in an array called **adapterNames** in the each **provider**. This can be used to display a rooted tree view 

### Testing Instructions
- Login to backend
- Visit http://localhost/media-manager-improvement/administrator/index.php?option=com_media&format=json&task=api.files&path=local-0:/

### Expected result
- You should see default files in the images folder as usual
- Try any other one like http://localhost/media-manager-improvement/administrator/index.php?option=com_media&format=json&task=api.files&path=local-1110:/
- This should throw an error

**Note : This is a backend update, front end need to be fixed to use this rooted tree view**

**Note : Follow up PR is coming for UI/UX fix**


administrator/components/com_media/Model/Api.php
```diff
@@ -97,9 +104,10 @@ private function getAdapter($name)
 	 * can be found in AdapterInterface::getFile().
 	 *
 	 * @param   string  $adapter  The adapter
+	 * @param   string  $account  The account for adapter
```
laoneo: Is this neede

administrator/components/com_media/Controller/Api.php
```diff
@@ -108,16 +107,16 @@ public function files()
 					// Grab options
 					$options = array();
 					$options['url'] = $this->input->getBool('url', false);
-
 					$data = $this->getModel()->getFiles($adapter, $path, $this->input->getWord('filter'), $options);
-
 					break;
+
```
laoneo, line 112: Why these empty lines?

laoneo (changes requested):

administrator/components/com_media/Controller/Api.php
kasvith: it is easy for anyone for read the file, if we have it like

```
case case1:
	break;
case case2:
	break;

```
it seems harder for read them because of everything is together

PR #282, opened 2017-08-23
UI/UX fix for Media Manager to enable Multiple account architecture

kasvith:
After #281 the Media Manager Client was broken because of a break in the backend. This PR will fix this issue allowing new features in the frontend

### Summary of Changes
- Added rooted provider menus in the tree
image

### Testing Instructions
- This should display any adapter with their rooted accounts

### Changes required
- We need more tree styling
- We need more animations for this


administrator/components/com_media/resources/scripts/store/state.js
```diff
@@ -7,20 +7,32 @@ if (options.providers === undefined || options.providers.length === 0) {
...
+
+            result.accounts.push(account);
+        }
+
+        return result;
     }),
     // The loaded directories
     directories: options.providers.map((disk) => {
-        return {path: disk.name + ':/', name: disk.displayName, directories: [], files: [], directory: null}
+        return {path: disk.name + '-0:/', name: disk.displayName, directories: [], files: [], directory: null}
```
laoneo: Must this be a hardcoded `0`?

kasvith: yup, we used default provider before, now we have default account

kasvith: Since we agreed to load default one based on the Plugin order we use this.

laoneo: What happens when there is no account with 0?

kasvith: When MM boots up no provider will be loaded automatically(because default one is missing), but when you click manually they will work, we still have not displaying errors in the MM.
After we do we can display an error. It means no adapter is associated with certain provider.

kasvith: Error handling and displaying still missing

laoneo: Do you have the list of adapters available? If yes, just take the first one.

kasvith: yes. We are taking 1st adapter and 1st account associated with it, if there are no any account that needed to be handled

laoneo: But should you not take then the path from the adapter instead of hardcoding it?

kasvith: even if we get it will be something like dropbox-0 for 1st account.

we already have 1st adapter as dropbox and we take it's 1st account by adding 0

kasvith: disk.name is already set to the 1st provider

administrator/components/com_media/resources/scripts/store/state.js
```diff
@@ -7,20 +7,32 @@ if (options.providers === undefined || options.providers.length === 0) {
...
+
+        for(let i = 0; i < provider.adapterNames.length; i++)
+        {
+            let account = {
+                root : provider.name + '-' + i + ':/',
+                displayName : provider.adapterNames[i],
+            };
+
+            result.accounts.push(account);
+        }
```
dneukirchen: You can simplify this using `Array.prototype.map()` (3-4 lines instead of 10).

administrator/components/com_media/resources/scripts/mediamanager.js
```diff
@@ -20,6 +21,7 @@ import store from './store/store';
 Vue.use(Translate);
 
 // Register the vue components
+Vue.component('media-adapter', Adapter);
```
dneukirchen: Not so happy with the naming "adapter". 
An adapter is just a wrapper for the media provider api.
We should not use that in any part of the frontend. We can discuss naming in our next meeting.

administrator/components/com_media/resources/scripts/store/state.js
dneukirchen: Also not very with the hardcoding here. Please use the path from the first adapter.

dneukirchen (changes requested):

administrator/components/com_media/resources/scripts/store/state.js
kasvith: it is path from 1st adapter

kasvith: @dneukirchen the XHR request is like local-0:/path/to/file. So it needs two things

- adapter
- account index

what i do in here is obtaining 1st adapter and selecting 1st account index associated within it.

administrator/components/com_media/resources/scripts/mediamanager.js
kasvith: do you have any naming suggestion?

administrator/components/com_media/resources/scripts/store/state.js
kasvith: will try my js knowledge is poor btw

administrator/components/com_media/resources/scripts/mediamanager.js
dneukirchen: I prefer disk and drivers.

Driver is a representation of a media provider which can have multiple disks and internally uses the Adapter to talk to the provider.

kasvith: ok we can do it, it is just a rename

administrator/components/com_media/resources/scripts/store/state.js
dneukirchen: Yes, but we should not hardcode that in. This should be passed to the client from the server. Frontend does not need to know that much about how the server handles multiple accounts.

kasvith: so in this case i suppose we need to use the disks variable defined before instead if remapping, but do we really need it?

dneukirchen:
Perhaps we should not group disks (currently named adapters) by drivers (currently named providers).

I see 2 problems with that:

1) If user uses only 1 disk of a certain provider, it makes no sense to me to have that additional tree node in the frontend.
2) I think that it could be nice to make the disk sortable, so that the user can decide in which order they appear in frontend.

kasvith:
1. We can remove a node if there is single account associated with it(in future, this PR only fix for client to reflect changes in backend)

2. It is nice, but grouping does not have a conflict with it

kasvith:
My suggestion. Instead of letting users sort drives, we should let them mark their favourite drives
This makes user's life much more easier.

kasvith:
Having a single account per disk is totally ok with me. For example Linux has only root and it is still there in Dolphin

Even in Windows explorar, if you only have single network drive, Network Drive node still remains

administrator/components/com_media/resources/scripts/store/state.js
kasvith: its done. can you check

dneukirchen: lgtm

kasvith:
#283 describe the idea

laoneo:
I merged it to move on, we can change stuff later again.

laoneo:
Ty kasun!

kasvith:
Tj ? :o

PR #284, opened 2017-08-26
Local Adapter with customizable Folder list

kasvith:
### Summary of Changes
This is a fellow up PR for #282 
Please resolve it before proceed.

- Updated Local Adapter to use a folder list
image

- Updated Translations for new changes

### Testing Instructions
- Go to Filesystem - Local plugin
- Under directories add any directory  you want
- Save
- Go to Media Manager

### Expected result
- Added directories should be mounted


plugins/filesystem/local/local.xml
```diff
@@ -29,6 +31,19 @@
...
+					name="directories"
+					type="subform"
+					label="PLG_FILESYSTEM_LOCAL_DIRECTORIES_LABEL"
+					description="PLG_FILESYSTEM_LOCAL_DIRECTORIES_DESC"
+					multiple="true"
+					layout="joomla.form.field.subform.repeatable"
+					buttons="add,remove,move"
+				>
+					<form>
+						<field name="directory" type="folderlist" default="images" label="Directory" filter="" exclude="" stripext="" />
```
laoneo: String needs to be translated.

plugins/filesystem/local/local.xml
```diff
@@ -29,6 +31,19 @@
 					description="PLG_FILESYSTEM_LOCAL_DISPLAY_NAME_DESC"
 					default="Local"
 				/>
+				<field
+					name="directories"
+					type="subform"
+					label="PLG_FILESYSTEM_LOCAL_DIRECTORIES_LABEL"
+					description="PLG_FILESYSTEM_LOCAL_DIRECTORIES_DESC"
+					multiple="true"
+					layout="joomla.form.field.subform.repeatable"
```
laoneo: Probably the table layout is nicer

plugins/filesystem/local/local.xml
```diff
@@ -12,6 +12,8 @@
 	<files>
 		<filename plugin="local">local.php</filename>
 		<folder>Adapter</folder>
+		<folder>forms</folder>
+		<filename>forms/directory_list_form.xml</filename>
```
laoneo: Is this needed?

plugins/filesystem/local/local.php
```diff
@@ -37,11 +37,26 @@ class PlgFileSystemLocal extends CMSPlugin
...
 	public function onFileSystemGetAdapters()
 	{
-		// Compile the root path
+		$adapters = [];
+
+		// Compile the root path and load default images folder
 		$filePath = ComponentHelper::getParams('com_media')->get('file_path', 'images');
 		$root = JPATH_ROOT . '/' . $filePath;
 		$root = rtrim($root) . '/';
+		$adapters[] = new \Joomla\Plugin\Filesystem\Local\Adapter\LocalAdapter($root, $filePath);
```
laoneo: Add images as default value instead of hardcoding it, perhaps somebody doesn't want to have images.

laoneo (changes requested):

plugins/filesystem/local/local.xml
kasvith: old artifacts lol

plugins/filesystem/local/local.xml
kasvith: done

plugins/filesystem/local/local.xml
kasvith: done

plugins/filesystem/local/local.php
kasvith: so shall we remove option from com_media ? we get images from there

laoneo: yes

plugins/filesystem/local/local.php
```diff
@@ -37,11 +37,27 @@ class PlgFileSystemLocal extends CMSPlugin
 	 */
 	public function onFileSystemGetAdapters()
 	{
-		// Compile the root path
-		$filePath = ComponentHelper::getParams('com_media')->get('file_path', 'images');
-		$root = JPATH_ROOT . '/' . $filePath;
-		$root = rtrim($root) . '/';
+		$adapters = [];
+		$directories = $this->params->get('directories', []);
```
laoneo: The default value must be here also images, when not set to be consistent with the UI.

kasvith: The return value from subform is a json string and we parse it below. I'll fix it

laoneo:
Thanks.

PR #288, opened 2017-08-28
Media field 

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Custom elements with some jQuery crappines till we get the modal, tooltip and popover as custom elements as well.

@laoneo @kasvith let me know what we need more here


kasvith:
thanks @dgt41

PR #290, opened 2017-08-29
Milestone1 media field retake

dgrammatiko:
Pull Request for Issue # .


administrator/components/com_media/resources/scripts/components/browser/items/item.js
```diff
@@ -42,9 +42,29 @@ export default {
          * @param event
          */
         function handleClick(event) {
-            var e = createEvent('onMediaFileSelected');
-            e.item = item;
-            window.parent.document.dispatchEvent(e);
+            const rootPath = Joomla.getOptions('com_media').fileBaseRelativeUrl;
+            const cloudRootPath = Joomla.getOptions('com_media').fileBaseRelativeUrl; //@todo return the cloud root...
+            const isCloud = false; //@todo return true if file is on the cloud
```
kasvith: this is not need, we treat every adapter in the same way, either you used local or whatever path is resolved by the API call

administrator/components/com_media/resources/scripts/components/browser/items/item.js
```diff
@@ -42,9 +42,29 @@ export default {
...
+	        let path = false;
+
+            if (item.type === 'file') {
+                if (isCloud) {
```
kasvith: applies above review

administrator/components/com_media/tmpl/media/default.php
```diff
@@ -40,6 +40,7 @@
 	'csrfToken'               => JSession::getFormToken(),
 	'filePath'                => $params->get('file_path', 'images'),
 	'fileBaseUrl'             => JUri::root() . $params->get('file_path', 'images'),
+	'fileBaseRelativeUrl'     => $params->get('file_path', 'images'),
```
kasvith: we dont use this param anymore. will remove it
So dont use this as we can add multiple folders to local adapter too

kasvith (changes requested):

administrator/components/com_media/resources/scripts/components/browser/items/item.js
dgrammatiko: where is this info propagated in the front end?

kasvith: it is in the options, but you don't need to separate clouds and non clouds
we treat all in the same way. Also user can have multiple folders in Local Adapter

dgrammatiko: @kasvith don't mix PHP with JS, what you are talking about is PHP, what is needed here in the JS scope is a var with the root path (the first part that will form the relative or absolute path). Is such var stored somewhere? If not it needs to be. Best way is to attach this info to the property of the item

kasvith: no. it is not. 
actually there is no fileBaseRelativeUrl anymore. I will remove it as we can have multiple folders in MM

administrator/components/com_media/tmpl/media/default.php
kasvith: @dgt41  here

administrator/components/com_media/resources/scripts/components/browser/items/item.js
```diff
@@ -37,14 +37,63 @@ export default {
...
+	            Joomla.request({
+		            url: `${apiBaseUrl}&task=api.files&url=true&path=${item.path}&${csrf}=1`,
+		            method: 'GET',
+		            perform: true,
+		            headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+		            onSuccess: (response) => {
+			            const resp = JSON.parse(response);
+			            if (resp.success === true) {
+				            if (resp.data[0].url) {
+					            if (/local-/.test(item.path)) {
```
kasvith, line 81: we can have infinite root folders for image storages.Also default **images** folder can be removed by a user from settings

kasvith (changes requested):

administrator/components/com_media/resources/scripts/components/browser/items/item.js
dgrammatiko: I am not checking against `local-0` so the code will happily execute also for `local-999999999999`

dgrammatiko: @kasvith is it posible to make `thumb_path ` camel case, e.g. `thumbPath` ?

kasvith: Sure can do

laoneo:
Tested it and it worked on images for articles

administrator/components/com_media/resources/scripts/components/browser/items/item.js
kasvith: done in #292

administrator/components/com_media/resources/scripts/components/browser/items/item.js
```diff
@@ -37,14 +37,63 @@ export default {
...
+	     */
+        function sendEvent(data) {
+	        const ev = new CustomEvent('onMediaFileSelected', {"bubbles":true, "cancelable":false, "detail": data});
+	        window.parent.document.dispatchEvent(ev);
+        }
         /**
          * Handle the click event
          * @param event
          */
         function handleClick(event) {
```
kasvith, line 60: We want to send the XHR when user Selects an supported file type and then when **he clicks the Select button**.
Not while he clicking every file on the Media Browser. Otherwise you will end up making your every file in a cloud being shareable which is a bad flaw

kasvith (changes requested):

kasvith:
Not working for **Non Local** file systems.
Please update to latest code, it fixes a bug in dropbox adapter.

I did it my self and i assure XHRs are OK

administrator/components/com_media/resources/scripts/components/browser/items/item.js
kasvith: Closed #292

PR #291, opened 2017-08-29
removed old unused params in the core

kasvith:
Pull Request for Issue # .

### Summary of Changes
Removed old unused parameters from media manager

### DIFF
--- a/administrator/components/com_media/config.xml
+++ b/administrator/components/com_media/config.xml
@@ -31,6 +31,24 @@
 		/>
 
 		<field
+			name="file_path"
+			type="text"
+			label="COM_MEDIA_FIELD_PATH_FILE_FOLDER_LABEL"
+			description="COM_MEDIA_FIELD_PATH_FILE_FOLDER_DESC"
+			size="50"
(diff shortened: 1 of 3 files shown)


administrator/components/com_media/config.xml
```diff
@@ -31,24 +31,6 @@
 		/>
 
 		<field
-			name="file_path"
```
laoneo: Because of backwards compatibility we can't remove the options here. They need to stay as default. It will take at least till Joomla 5 till we can remove the options.

laoneo (changes requested):

laoneo:
Does it still work without the parameters? Does the client not fail?

kasvith:
no client works

administrator/components/com_media/config.xml
kasvith: is this only for file_path ?

laoneo: No both options.

laoneo:
Thanks.

PR #308, opened 2017-09-20
Use only one icon for the listview layout toggle

dneukirchen:
bildschirmfoto 2017-09-20 um 21 44 48
bildschirmfoto 2017-09-20 um 21 44 54


kasvith:
Why don't we use separate icons ?

dneukirchen:
We dont need both of them at the same time. 
The icon changes depending on the selected view layout.

PR #389, opened 2017-09-27
Remove description because the tooltip got removed

laoneo:
### Summary of Changes
As there are no tooltips for forms anymore, the local adapter description looks awkward.

### Testing Instructions
Open the local filesystem plugin.

### Expected result
image

### Actual result
image


dneukirchen:
Is this the recommended solution for every joomla extension? Imo we should keep the desc until a final decision/solution exists.

kasvith (commented):
I would agree to @dneukirchen in here

Issue #184, opened 2017-04-13
File uploading with Duplicate names


kasvith:
### Steps to reproduce the issue
- Go to Media Manager
- Upload a file named **abc.jpg** to any folder (just an example)
- Pick another file, name it also as **abc.jpg**
- Try to upload to same folder

### Expected result
- Automatically renamed new file in the folder or if the content identical, an error message it already exists in the current folder

### Actual result
- Media Manager stays silent in duplicate name uploads

### System information (as much as possible)
PHP 7.1 Window 10

### Additional comments
- We need a mechanism for rename files like **abc - (1).jpg** or etc

dneukirchen:
Thx for reporting. We had a discussion about this some days ago:

>David Neukirchen
So one question that popped up: how should we handle file upload requests, if the file already exists in the target directory?

> Dimitris Grammatiko
the old way was to refuse the upload
I suggest to create a filename-1.ext similar to articles with the same alias

>David Neukirchen
that would be possible... but maybe there are use-cases where the user wants to replace a file with an updated version of that file? I dont know...

> Dimitris Grammatiko shared a link
In the old controller and plugins I had introduced a way to allow overwriting: https://github.com/joomla-projects/media-manager-improvement/pull/19/files#diff-e5953bc3eadad37980d677840c9d4f6dR154

> Thomas Hunziker
We had that discussion in the past. Just refuse the upload or if you can ask if it should overwrite the file. Don't change the filename to something else, that would be very unexpected.

We decided to not override or rename the file automatically. We need to ask the user if he wants to override the file, otherwise he needs to change the filename.

kasvith:
@dneukirchen I suggest to add an option which will automatically rename duplicate files or ignore them in options(should be in the Media Manager itself), so user can toggle it anytime, if we plan to do drag n drop uploads, user will be annoyed to each time look if file exists or rename files

kasvith:
@dneukirchen if user want to update an file, we can do a simple and handy thing, we can create a context menu, where user can update the selected file

dneukirchen:
Good ideas!

Config Options are kind of handicapped in the joomla world. Imo joomla has already too many of them. We need to be very careful and think twice, if we want to add another one. In this case i would say we should stick with a dialog/context modal like you suggested in you second comment.

So the workflow for already existing files could be:

* Ask the user if he wants to override it
     * yes => override the file (send an additional param to api)
     * no => show input field where the user can change the filename manually

[...]

PR #391, opened 2017-09-27
Feature/improve file upload with duplicate file name

dneukirchen:
Pull Request for Issue #184 .

bildschirmfoto 2017-09-27 um 16 52 43

When the file already exists, the user has to answer if he wants to override it. 

bildschirmfoto 2017-09-27 um 16 53 12

@laoneo we need to implement that on server side:
when we send a "override=true" param to server along with an upload request, the server should override the existing file. If override=true is not present in request params and the file already exists, the server should return a 409 error response and not override or modify the file. The rest will be done in the client..


dneukirchen:
If override=true is not present in request params and the file already exists, the server should return a 409 error response and not override or modify the file. The rest will be done in the client.

kasvith:
@dneukirchen shouldn't it be override=1 ?

dneukirchen:
only if we use getBool() on server side, which we dont.

kasvith:
gj @dneukirchen

PR #394, opened 2017-09-27
Suggestion: move checkall to upper left

dneukirchen:
What do you think about moving the checkAll checkbox in the upper left corner?

Before:
bildschirmfoto 2017-09-27 um 20 45 49

After: 
bildschirmfoto 2017-09-27 um 20 44 59


kasvith (commented):
I would like to see it on toolbar

ciar4n:
This works for me.

The only issue is it might be associated with the breadcrumbs. Top left is the same location as android however they hide the breadcrumbs when one or more items are selected and replace it with a select all checkbox followed by the number of items selected. This resolves the breadcrumbs association issue.

ciar4n (approved):

rgmears:
This is a variation on what @dneukirchen has presented. I admit to changing the colour scheme a little, but the drift here is to make the work area as large as possible:
j4-media-manager-1

dneukirchen:
thx @rgmears i like the idea, but unfortunately we cant change the joomla admin toolbar. This is something the joomla 4 template team is currently working on.

laoneo:
I'm here on the same page as @dneukirchen, this should be discussed with the template team and then we will adapt to their standard. But I think it is worth to rethink the toolbar setup as what we have now is the same since ages.

For now I would like to keep focus on the issue the pr tries to solve, to move the select all checkbox to the left side.

rgmears:
I'm curious @dneukirchen and @laoneo why the breadcrumbs are needed at all. It's quite obvious where the user is in terms of folder hierarchy. If the breadcrumbs were omitted and the select all checkbox returned to its original location the images pane could be even taller (almost the full height of the screen).
j4-media-manager-4

brianteeman:
They are essential for accessibility!

rgmears:
Why?

brianteeman:
https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-location.html

rgmears:
Seems redundant to me @brianteeman. 
There is no breadcrumb trail on any other screen in the backend.

brianteeman:
No other screen in the admin allows navigation in this way

rgmears:
Well navigation this way -- breadcrumbs -- only allows for going backwards and is IMO useless. Just a waste of screen space.

brianteeman:
You dont understand the accessibility - it is not about navigation anywhere - it is about knowing where you are

brianteeman:
It is only useless to you because you do not face any barriers when you use the internet

rgmears:
Seems pretty obvious to me where I am in the folder hierarchy.
j4-media-manager-detail-1
Although I don't understand what you mean by facing barriers. What sort of barriers?

brianteeman:
1. you can see
2. you have cognitive skills that allow you to navigate complex paths and understand them

rgmears:
Why would blind people be working with images?
And, if the breadcrumbs are only about knowing where you are, why are they linked?
Do people lacking "cognitive skills" add content to Joomla! websites?

brianteeman:
OK I am not even going to comment on your totally ignorant and offensive comments. I hope that you will spend a few moments to think and reflect about what you just said.

rgmears:
I think it would be more appropriate for you to answer my questions.

brianteeman:
Sorry but I dont respond to ignorant offensive comments

mbabker:
> Do people lacking "cognitive skills" add content to Joomla! websites?

Yes.  People with physical and mental disabilities (as in having impaired "cognitive skills") are commonly managing websites (not just Joomla installations).

rgmears:
> People with physical and mental disabilities (as in having impaired "cognitive skills") are commonly managing websites

OK. But, how do the breadcrumbs help such people?

mbabker:
A breadcrumb tree more capably communicates the current navigation path than trying to determine it from the menu structure applicable to that area (in the media manager that'd be the folder tree in the sidebar, in your site frontend that's usually your main menu tree).  It also allows users to easily traverse up the site navigation without having to find the content in the menu tree.

rgmears:
I notice (now) that Windows Explorer has an address bar which emulates breadcrumbs. I've never used it. And Nemo (files browser on Linux) has a top bar which performs the same function as breadcrumbs, which I glance at occasionally. Still, to go to another folder one has to scroll the folder tree in sidebars on both Windows and Linux, unless, of course, one is just going back a folder or two.

It still seems superfluous to me to have breadcrumbs on the Media Screen.

dneukirchen:
Agree with @mbabker and @brianteeman. We should keep the breadcrumb. For many users it is still a very important and easy to use navigation and orientation element. 

@rgmears i really appreciate your interest and efford and we are always open minded for new ideas and suggestions. 

For this PR: lets get back to the topic. Dont hesitate to open separate issues or send a pull request if you have any further feedback, suggestions or issues with/for the new media manager.

brianteeman:
Let me be clear. breadcrumbs are NOT about navigation. They are about telling you where you are. Breadcrumbs do not have to be clickable, although typically they are, that is just a bonus.

rgmears:
With the folder tree being so prominent the breadcrumbs, in this case, are a distraction.

dneukirchen:
guys ;-) this PR is not about what a breadcrumb is (not) or should be... 

https://github.com/joomla-projects/media-manager-improvement/issues/new if you really want to discuss that any further...

rgmears:
Done: https://github.com/joomla-projects/media-manager-improvement/issues/424

brianteeman:
I tend to agree with @ciar4n 
> The only issue is it might be associated with the breadcrumbs.

laoneo:
As far as I got the impression is that most here do agree that it would make sense to move the select all action to the left side. So I suggest that when we get rid of the >, that it doesn't look like it belongs to the breadcrumbs, we can merge it.

dneukirchen:
made adjustements

bildschirmfoto 2017-09-29 um 18 49 43

laoneo:
Thanks

rgmears:
Just out of curiosity, what am I as a user likely to do after I've selected ALL of the images?

dneukirchen:
delete

rgmears:
And where on the page do I do that?

dneukirchen:
There is a red button in the toolbar.

In the future i can think of other batch operations. Something like download as zip or moving multiple items between folders.

rgmears:
What are the **-** and **+** buttons for?

dneukirchen:
they in/decrease the grid item size. 
probably we need to change those buttons to somthing like http://fontawesome.io/icon/search-plus/

rgmears:
OK. Thanks.

rgmears:
@brianteeman 
Regarding the breadcrumbs, I noticed that it is built into the code for the Media Manager page. If it were placed as a module (as other admin template elements are) then it would be the user's choice whether to display them or not. Of course, though, that would mean moving the checkall back to where it was originally.

rgmears:
I think it would be better to move the check-all button back to where it was. Then it is beside the "-" button which is the next action in the sequence. One would check (select) all then delete from what has been said here.

PR #405, opened 2017-09-28
Feed that hound...

ciar4n:
Pull Request for Issue # .


ciar4n:
Hound here only lints rules that have been changed so there may be still a few errors left. That is most of them for now.

dneukirchen:
thx @ciar4n

Issue #416, opened 2017-09-28
Only work with media files


brianteeman:
**This is really bad.** It can open up the entire web site to anyone with media manager access. Remember that can include low level users. Presumably (although currently broken) the media manager will be used on the front end as well. Making this even more dangerous.

 As can be seen below I have opened up the site to allow anyone to delete or rename any file - including php - on the site

<img width="468" alt="screenshotr15-06-10" src="https://user-images.githubusercontent.com/1296369/30971998-500ae1f2-a461-11e7-81a8-6df801199271.png">
<img width="457" alt="screenshotr15-07-05" src="https://user-images.githubusercontent.com/1296369/30971997-5000555c-a461-11e7-90b3-824c3ae7fce6.png">
<img width="393" alt="screenshotr15-06-26" src="https://user-images.githubusercontent.com/1296369/30971996-4fe9d05c-a461-11e7-813f-f78721e29048.png">

brianteeman:
@dneukirchen  this is a critical issue and it cannot be merged without this being fixed

kasvith:
How anyone without admin rights can use media manager?

dneukirchen:
Yes, thx @brianteeman 
We will resolve this asap. Should be an easy fix.

brianteeman:
@kasvith at the front end when creating an article. Also a low level admin user with just access to create content in the admin

kasvith:
We will add user level permissions to NMM later, for now we would just filter the *.php* files from the rest

brianteeman:
NO! It is nothing to do with permissions. That is not the correct way. You should only be displaying defined media files. You should not be hardcoding anything to filter out files. Simply filtering out *.php is nowhere near enough. What about js etc. 

Please look at the current media manager. It only displays files that are set in the options for Legal Extensions (File Types). NMM is ignoring this setting and this is the correct way to do this - a user controlled WHITELIST of allowed files

laoneo:
If you put a .php file into the images folder in the old media manager, then it does show this file. So we are not badder as the old one. But we want to move forward, #419 should only serve and work with files which are configured as allowed file types in the settings.

brianteeman:
> If you put a .php file into the images folder in the old media manager, then it does show this file. So we are not badder as the old one

The problem was that the new local storage option exposed the entire folder tree and users are stupid.

Going to close this as #419 resolves this

PR #419, opened 2017-09-29
Only work with file extensions which are allowed

laoneo:
Pull Request for Issue #416.

### Summary of Changes
The media manager should only work with files which are allowed file extensions in the settings.
image

### Testing Instructions
Put a PHP or JS file in the images folder.

### Expected result
The none media files should not being displayed.

### Actual result
All files are displayed.

### DIFF
--- a/administrator/components/com_media/Model/ApiModel.php
+++ b/administrator/components/com_media/Model/ApiModel.php
@@ -143,15 +143,17 @@ class ApiModel extends BaseModel
 		}
 
 		// Add adapter prefix to all the files to be returned
-		foreach ($files as $file)
+		foreach ($files as $key => $file)
 		{
+			// Check if the file is valid
 			if ($file->type == 'file' && !$this->isMediaFile($file->path))
 			{
+				// Remove the file from the data
+				unset($files[$key]);
 				continue;
 			}
 
-			// If requested add options
-			// Url can be provided for a file
+			// Check if we need more information
 			if (isset($options['url']) && $options['url'] && $file->type == 'file')
 			{
 				if (isset($options['temp']) && $options['temp'])
@@ -168,7 +170,8 @@ class ApiModel extends BaseModel
 			$file->adapter = $adapter;
 		}
 
-		return $files;
+		// Return array with proper indexes
+		return array_values($files);
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


brianteeman:
Tested this successfully!!! The NMM now obeys the settings in the Legal Extensions options

administrator/components/com_media/Model/ApiModel.php
```diff
@@ -337,10 +385,45 @@ public function search($adapter, $path = '/', $needle, $recursive = true)
...
+			$this->allowedExtensions = ComponentHelper::getParams('com_media')->get(
+				'upload_extensions',
+				'bmp,csv,doc,gif,ico,jpg,jpeg,odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,BMP,CSV,DOC,GIF,ICO,JPG,JPEG,ODG,ODP,ODS,ODT,PDF,PNG,PPT,TXT,XCF,XLS'
+			);
+
+			$this->allowedExtensions = explode(',', $this->allowedExtensions);
+		}
+
+		$extension = substr($path, strrpos($path, '.') + 1);
+		return in_array($extension, $this->allowedExtensions);
```
dneukirchen: CS?

dneukirchen (commented):
Shouldnt we also filter in the local filesystem plugin?

something like

```
    $filter = "^.*\.(" . implode('|',$allowedExtensions) .")$";
    $files = JFolder::files($imgsPath, $filter);
```

administrator/components/com_media/Model/ApiModel.php
```diff
@@ -129,6 +145,11 @@ public function getFiles($adapter, $path = '/', $options = array())
 		// Add adapter prefix to all the files to be returned
 		foreach ($files as $file)
 		{
+			if ($file->type == 'file' && !$this->isMediaFile($file->path))
+			{
+				continue;
```
dneukirchen: This will not hide the files in the json response! They are still returned.
We should unset them or not get them from filesystem (see comment above).

dneukirchen (changes requested):

administrator/components/com_media/Model/ApiModel.php
laoneo: What for cs?

laoneo:
The plugin should not use the global options from com_media.

administrator/components/com_media/Model/ApiModel.php
dneukirchen: ;-) blank line before return statement?

dneukirchen (approved):

dneukirchen:
thx!

Issue #417, opened 2017-09-28
front end broken


brianteeman:
Testing on the front end and I always get

`Warning: require_once(C:\Apache24\htdocs\cms4\administrator/components/com_media/media.php): failed to open stream: No such file or directory in C:\Apache24\htdocs\cms4\components\com_media\media.php on line 18`

laoneo:
Please test #429, it should fix the issue.

brianteeman:
I would test it but the branch has a 500 error on the front end

500 Expression #1 of ORDER BY clause is not in SELECT list, references column 'nmmtest1.c.lft' which is not in SELECT list; this is incompatible with DISTINCT

laoneo:
Did you do a fresh installation?

brianteeman:
yes

laoneo:
I didn't do a new installation, if there is such an issue, then we have it probably in 4.0-dev as well as we didn't change something on that part.

brianteeman:
Sorry - yes you are correct - its broken upstream

PR #429, opened 2017-10-03
Frontend fix

laoneo:
Pull Request for Issue #417.

### Summary of Changes
Fixes the front end access to the media manager.


brianteeman:
Thanks - its a partial fix - the errors are now gone

There are two different displays of the modal in the front end - i cant see why but it seems to think that i am changing an image when i am on the images and links tab when there isnt an image there and its hard to tell from the pictures but the modals are very different sizes

### xtd-editor button

screenshotr17-24-51



### images and link tab
screenshotr17-24-34

laoneo:
We target for the next milestone to brush up the media form field and the xtd button. Important for now is that they are working on the front as well.

dneukirchen:
Agree, lets improve the frontend experience within the next milestone.

PR #431, opened 2017-10-04
Camel case tasks and unify sending the response

laoneo:
Camel case tasks and unify sending the response.

### DIFF
--- a/administrator/components/com_media/Controller/ApiController.php
+++ b/administrator/components/com_media/Controller/ApiController.php
@@ -63,7 +63,7 @@ class ApiController extends BaseController
 
 			if (!in_array($this->doTask, $this->taskMap))
 			{
-				throw new \Exception(\JText::sprintf('JLIB_APPLICATION_ERROR_TASK_NOT_FOUND', $task), 404);
+				throw new \Exception(\JText::sprintf('JLIB_APPLICATION_ERROR_TASK_NOT_FOUND', $task), 405);
 			}
 
 			$data = $this->$doTask();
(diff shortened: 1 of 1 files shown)


administrator/components/com_media/Controller/ApiController.php
```diff
@@ -56,24 +56,20 @@ public function execute($task)
...
+			$doTask = strtolower($method) . ucfirst($task);
 
 			// Record the actual task being fired
 			$this->doTask = $doTask;
 
-			if (in_array($this->doTask, $this->taskMap))
+			if (!in_array($this->doTask, $this->taskMap))
 			{
-				return $this->$doTask();
+				throw new \Exception(\JText::sprintf('JLIB_APPLICATION_ERROR_TASK_NOT_FOUND', $task), 404);
```
dneukirchen: i would suggest status code 405 instead of 404 to be more explicit.

administrator/components/com_media/Controller/ApiController.php
```diff
@@ -156,20 +149,16 @@ public function get_files()
 	 * 		index.php?option=com_media&task=api.files&path=/sampledata/fruitshop/test.jpg
 	 * 		/api/files/sampledata/fruitshop/test.jpg
 	 *
-	 * @return  void
+	 * @return  array  The data to send with the response
```
dneukirchen: it returns null

administrator/components/com_media/Controller/ApiController.php
dneukirchen: we can probably directly return the response instead of throwing and catching exception in the same method.

laoneo:
Done.

PR #446, opened 2017-10-23
 Adapter is responsible for filename normalization

laoneo:
The logic which file names are allowed should be handled and managed by the adapter and not the controller. This prevents us from situations on restrictive adapters when they do not allow certain file names that they can decide the filename and do not abort the file upload. The logic is changed the way that when a file or folder is created the file name is returned by the adapter of the new file or folder which get propagated to the client then. Similar as Google is doing it.

~Additionally files with spaces are allowed as I don't see a valid reason why it shouldn't work.~

### DIFF
--- a/administrator/components/com_media/Controller/ApiController.php
+++ b/administrator/components/com_media/Controller/ApiController.php
@@ -194,19 +194,17 @@ class ApiController extends BaseController
 		$mediaContent = base64_decode($content->get('content', '', 'raw'));
 		$override     = $content->get('override', false);
 
-		$name = $this->getSafeName($name);
-
 		if ($mediaContent)
 		{
-			$this->checkContent($name, $mediaContent);
+			$this->checkContent();
 
 			// A file needs to be created
-			$this->getModel()->createFile($adapter, $name, $path, $mediaContent, $override);
+			$name = $this->getModel()->createFile($adapter, $name, $path, $mediaContent, $override);
 		}
 		else
 		{
 			// A file needs to be created
-			$this->getModel()->createFolder($adapter, $name, $path, $override);
+			$name = $this->getModel()->createFolder($adapter, $name, $path, $override);
 		}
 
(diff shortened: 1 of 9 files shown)


administrator/components/com_media/Controller/ApiController.php
```diff
@@ -333,17 +331,14 @@ public function getModel($name = 'Api', $prefix = 'Administrator', $config = arr
 	}
 
 	/**
-	 * Performs various check if it is allowed to save the content with the given name.
-	 *
-	 * @param   string  $name          The filename
-	 * @param   string  $mediaContent  The media content
+	 * Performs various check if it is allowed to save the content.
```
brianteeman: Please correct the English to "checks"

kasvith:
Agree on @laoneo here, most clouds supports many kind of file names, even with spaces as they use a different method to store files

brianteeman:
Please correct Al the instances of normalize to normalise

kasvith (approved):
Nice work

wilsonge:
> Additionally files with spaces are allowed as I don't see a valid reason why it shouldn't work.

Can we do this in a separate PR. I think it's probably OK. But there's definitely some weirdness with how things get interpreted in web (because obviously in a real url spaces have to be replaced with `%20`. If it's in a separate URL it's much easier to revert if any problems show up later down the line

laoneo:
Changed the spelling errors and reverted the whitespace strip change. If this one gets merged I will do a followup with the support of whitespaces.

dneukirchen:
Is it possible to directly return the normalized item (instead of just the new name)?

laoneo:
You mean in the adapter? Because the API controller returns the new item already.
When done in the adapter, then it is mostly connected with a second request on cloud systems. If the full item is not needed, then we have here an increase of resources which are not necessary.

dneukirchen:
thx @laoneo

Issue #260, opened 2017-06-20
Normalize filenames - proposal


schnuti:
In issue #245 I argue that the characters allowed for filenames should be kept the same as with the old MM. This though do not imply that characters should be removed. Test Übergrößenträger.jpg as filename in the old MM.
I propose:
    
    JFilterOutput::stringURLSafe($string, $language = '')
       This method processes a string and replaces all accented UTF-8 characters by unaccented
       ASCII-7 "equivalents", whitespaces are replaced by hyphens and the string is lowercase.

1. Lowercase file names. I use the stringURLSafe method that results in allowed characters. Maybe this method has to be duplicated to a stringFileSafe method? The URL characters have nothing to do with the filesystem and could be changed in near future. As for the alias field in the DB tables a counter has to be added if filename already exists. (myfile-3.txt)  
EDIT: There is already a filter method stringUrlUnicodeSlug($string) for unicode slugs. 
2. Lowercase filename extensions. jpg, png, gif, bmp, txt, mpg a s o.
3. Replace jpeg with jpg and any other filetype with more than one extension. If not lowercase you get a LOT of different extensions. How do you like mylogo.pnG? I can use that on windows, can't upload with MM  --> - Error This file type is not supported. After manual upload MM handels it.   

Joomla can leave it to the users but why not help them with some normalizing.

ggppdk:
I have made a relevant PR here
https://github.com/joomla/joomla-cms/pull/12049

That does language transliteration for filenames
Also added a method for making filename unique 
-- **in a given folder**

About transliteration, it can be used, until support for both UTF-8 / UTF-16 is solidly implemented,

schnuti:
@ggppdk 
Great! I'll test it. I know that it works for me as I've used a method like that since Joomla 1.5. 
I saw that you allow upper case letters. Personally I hate camel case file names. When I then see a JPEG at the end, then I go offside. 
At the end of the day the file information should be stored in a DB table. Long name, description, alias, alt text, caption, credits and copyright, dates and a lot more as needed.    
Well Joomla always had problems with recommendations and best practices.

schnuti:
I  stumbled over a web page that very short describes why my normalizing is a good idea and the rules that should be followed. i.e I have some new friends ;)
[Michigan Technological University](http://www.mtu.edu/umc/services/digital/writing/characters-avoid/)  

I know that this is changing rapidly on the "top" level but Joomla has to be conservative.

schnuti:
[...]

PR #448, opened 2017-10-24
Allow spaces for file names on upload

laoneo:
Pull Request for Issue #260.

### Summary of Changes
Removes the code in the local adapter which strips out any none alpha numerich character. This is a follow up pr of #446.

### Testing Instructions
Upload a file with spaces.

### Expected result
Spaces do remain.

### Actual result
Spaces get stripped out.

### DIFF
--- a/plugins/filesystem/local/Adapter/LocalAdapter.php
+++ b/plugins/filesystem/local/Adapter/LocalAdapter.php
@@ -746,7 +746,6 @@ class LocalAdapter implements AdapterInterface
 		}
 
 		$nameWithoutExtension = substr($name, 0, strlen($name) - strlen($extension));
-		$nameWithoutExtension = preg_replace(array("/[\\s]/", '/[^a-zA-Z0-9_]/'), array('_', ''), $nameWithoutExtension);
 
 		return $nameWithoutExtension . $extension;
 	}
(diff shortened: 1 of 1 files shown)


schnuti:
This PR resolves what it is supposed to. It does not what the title says but reverts to the use of an old method. Allowing _,- and blanks.

Über-größen_ länge.jpg becomes ber-gren_ lnge.jpg

I have some new general comments on #260

schnuti:
Is it possible to override one specific method in LocalAdapter.php keeping the rest?
e.g. =>getSafeName($name)

laoneo:
Can you try again. The problem was that basename stripps out certain characters like the upper case Ü characters but accepts the lower case one.

schnuti:
Same result
Why shouldn't it? The characters get stripped at line 734

// Make the filename safe
   $name = \JFile::makeSafe($name);

If I comment out that line the AJAX response is
Message: COM_MEDIA_ERROR_UNABLE_TO_UPLOAD_FILE
Error:
0: File name must only contain alphanumeric characters and no spaces.

laoneo:
Ok, reverted to code to just fix to allow spaces and changed the title accordingly. Should be fine now. Will open another pr for the rest.

laoneo:
Lets discuss the umlaut issue in the pr #463.

schnuti:
To be more precise. It's not a fix for spaces. It's a fix for the missing dashes mentioned in 260.
Allowed is now _ (underscore), - (dash) and ' ' (space, blank, whitespace) beside numbers and "english" characters.
Never mind, it's ok as is for now.

Issue #462, opened 2017-10-27
Unicode filenames breaks MM


schnuti:
### Expected result
The file should appear in MM with "funny" characters as in J3.x
alt. skip files with unicoded names.

### Actual result
No files or folders at all are displayed after a short "progress". 

### System information (as much as possible)
XAMPP local test, Win10

schnuti:
This is because json_encode($array) throws an error when strings in the array is non UTF-8 as with Windows.
Solution 1:  json_encode($array, JSON_PARTIAL_OUTPUT_ON_ERROR) skips the non UTF-8 strings.
Solution 2: require PHP 7.1 for Joomla. That version handles the Windows filesystem.

laoneo:
#463 should fix this issue.

laoneo:
Issue can be closed as the pr got merged. Thanks for reporting.

schnuti:
@laoneo 
Did you really test with a Windows system and PHP version < 7.1 ? With a PHP version > 7.0 or Linux the issue is solved. I still havn't seen anything about that Joomla 4 will raise the requirement to PHP > version 7.0. 
As I store utf8 filenames in the DB of my local system, I require PHP > 7.0. i.e. I'm not able to make a quick test. Better for Joomla is to raise the requirement.

schnuti:
Well, I remembered, I still have an older XAMPP instance with PHP 7.0.4. After downloading the code from here I made a quick test. As I presumed - the problem remains. You have my two alternative solutions above. Image 1: fresh installation, Image 2: Übergrößen-länge.jpg added. = broken view. You can reopen the issue.

![original](https://user-images.githubusercontent.com/956537/34826103-89125b08-f6d5-11e7-8709-b4d204b7c284.jpg)

![unicodefile](https://user-images.githubusercontent.com/956537/34826116-9ae0a79a-f6d5-11e7-9bd4-e5b59a2b6315.jpg)

laoneo:
It works on Linux, hope somebody else can help here out.

kasvith:
I am using XAMMP with PHP 7, Windows 10 x64

I copied a file with unicode characters to the images directory and here is the result
![image](https://user-images.githubusercontent.com/13379595/34863567-1e49b53e-f796-11e7-920a-7ad32ba42eae.png)

In NMM

![image](https://user-images.githubusercontent.com/13379595/34863574-2b03b5f4-f796-11e7-8004-169f8eee68bd.png)

[...]

PR #463, opened 2017-10-31
Don't strip out special characters for display

laoneo:
Pull Request for comment https://github.com/joomla-projects/media-manager-improvement/pull/448#issuecomment-340257963.

### Summary of Changes
If a file is uploaded trough FTP with the name Über-größen_ länge.jpg, then it becomes ber-gren_ lnge.jpg. This is due some stripping in the basename function of PHP because of locale issues, this pr just stripps out the last part of the path.

### Testing Instructions
Upload an image with the name "Über-größen_ länge.jpg" trough FTP to the images directory.

### Expected result
Image is displayed with the name Über-größen_ länge.jpg.

### Actual result
Ü is stripped out from the image name Über-größen_ länge.jpg.

### DIFF
--- a/plugins/filesystem/local/Adapter/LocalAdapter.php
+++ b/plugins/filesystem/local/Adapter/LocalAdapter.php
@@ -308,7 +308,7 @@ class LocalAdapter implements AdapterInterface
 		// Set the values
 		$obj            = new \stdClass;
 		$obj->type      = $isDir ? 'dir' : 'file';
-		$obj->name      = basename($path);
+		$obj->name      = $this->getFileName($path);
 		$obj->path      = str_replace($this->rootPath, '/', $path);
 		$obj->localpath = $path;
 		$obj->extension = !$isDir ? \JFile::getExt($obj->name) : '';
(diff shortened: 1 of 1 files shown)


schnuti:
Is this the same as my issue #462?
Über-größen_ länge.jpg breaks the listing on Windows. i.e. Shows nothing. 
Is this the same fix as in 448? I did not check if it solved this issue,
BUT I saw that not only filenames where shown on screen but the whole path to the file/folder.

schnuti:
Windows is probably a "better" test environment for character issues. But be aware that the character use is a user setting on Linux, it doesn't have to be UTF8.

schnuti:
The problem remains. German characters lets the view fail completely. Don't worry, it's not only related to German ;)

Ãœb er_grÃ¶ÃŸen-lÃ¤nge as filename does not break the view. The jpg file image is of course not shown. This filename is stored as I deactivated the filename checks.  i.e. in bytes and bits it's utf8.

schnuti:
As far as I can see, there is no general solution. I found that for json_encode() "All string data must be UTF-8 encoded. " 
I reverted your basename changes as it added the path to the filenames and then
edited the code at line 311 in Controller/ApiController.php for a quick test.
This avoids to throw an exception and skips some of the data. The parameter 'JSON_UNESCAPED_UNICODE' did not do any difference for me. 
`
    // Send the data
    // echo new JsonResponse($data);
    // json_encode accepts utf-8 and some parameters

    echo json_encode($data, JSON_PARTIAL_OUTPUT_ON_ERROR);
`
Is now working on my local windows. Files with "extended" characters are missing but I do not miss them ;)

Issue #479, opened 2017-12-07
Make controls work in crop plugin


laoneo:
The controls in the crop plugin do actually not work.

PR #480, opened 2017-12-07
Make crop controls work

laoneo:
The controls in the crop plugin are actualy read only. This pr makes them working.

Pr for #479.


laoneo:
Merging it as I do a refactoring on resize.

PR #484, opened 2017-12-07
Fix up plugin UI

ciar4n:
Pull Request for Issue # .

### Before
image

image

### After
image

image


laoneo:
For me it shows the buttons like 
image

Once we even had them like 
image
 
Would it be possible to give them again some love?

ciar4n:
It seems that the radio field type has changed.. I think this is something that will have to be fixed in core first.

Issue #490, opened 2018-01-11
Bug: Width-height ratio on cropping


schnuti:
### Steps to reproduce the issue
1. Click on one of the radio buttons. The radio button do not get checked.
2. More important - If you once click on one of the radio buttons you are trapped. You can choose another ratio but not reset to "free" width/height. Another radio button or a reset button is needed. I think I would prefer a reset to original start values (images width and height).

laoneo:
True.

Please open for every problem one issue, it helps keeping the focus.

schnuti:
I think the two issues are that close related that they should stay together.

laoneo:
Can you please test #492 which should fix point 2.

@dgt41, @C-Lodder or @ciar4n any clue why the radio buttons do not work propperly? Something broken in core?

schnuti:
Part 2 has a PR 492 and an issue for part 1 has bin added.

PR #492, opened 2018-01-13
Add reset option for aspect ratio in crop plugin

laoneo:
Pull Request for Issue #490.

### Summary of Changes
Adds a reset button for the aspect ratio in the crop plugin.

image


brianteeman:
From the image thats just a radio button option and not a reset button. It should be a button. My 2c

schnuti:
Tested and it works!
@brianteeman It works as a Reset as long as you do not change to e.g. the resize view. Width and height go back to the original **in the crop view**.  You do a complete Reset with the button in the toolbar. This is ok for me.

@laoneo You were right. It was two issues. I'll add an issue for the "not checked" radio buttons so issue 490 can be closed.

brianteeman:
it might work but its still not a button as per the title

laoneo:
Changed the title.

dgrammatiko:
For the other issue let’s add a first option: original aspect ratio which should be selected by default

schnuti:
@dgt41 You mean yet another option I suppose. You still have to have an option where you can set width/height/pos-x/pos-y without any aspect ratio restrictions.

ciar4n:
@brianteeman For these to appear as a button group we wait for https://github.com/joomla/joomla-cms/pull/19326 to be merged.

Issue #495, opened 2018-01-14
Distinct values for image rotate


schnuti:
### Steps to reproduce the issue
If you rotate an image to other values than 90, 180, 270 degrees the image gets cut off. 

### Proposal
A canvas resize to fit the whole image. I know it's harder to achive but you could then make some minor rotation without cutting off the image.
Alternative: Use radio buttons for the distinct values 90, 180, 270, (360) degrees. I see that 90 and 270 degrees are handled correct.
Alternative 2: Keep the free rotation AND add distinct values with radio buttons.

ciar4n:
For me it would make sense to have the resize and rotate on the same panel as you would rarely do one without the other. Slightly off topic and probably not possible.

Example... most likely a user would zoom this image to remove the black corners. 

![image](https://user-images.githubusercontent.com/2803503/35038514-f10254ce-fb72-11e7-8c75-43795c41e7a7.png)

ciar4n:
@schnuti I think what you suggest is a good idea. My preference would be keep the free rotation AND add distinct values with radio buttons.

schnuti:
@ciar4n Please have a look at PR #496 for testing this. I kept free rotation but with a number field type where you can type in the degrees **OR** use the steps. I skipped the slider.

I do not think you can fix the black corners by zooming the image as the zoom is done including the black corners (the canvas). You could crop the image to get rid of the black corners if you didn't rotate to much. That's why I say that it should be possible to enlarge the image when resizing.  Rotate - crop to remove black corners - enlarge to original or wanted size. Am I right? I only believe so, not quite sure. 

When merging my patches on GitHub my change of the image quality fields to number field types got included (not planed)
You normally know what you want to use, the default 80 or 90, 75, 70 or whatever and you enter it directly. You do not want to open a long list with numbers from 1 to 100 (integer field type)

ciar4n:
Sorry you are right.. I meant to say crop not resize.

schnuti:
I found that the workflow I described works as is.
You are able to enlarge the image. Enter the width or height in the field. DO NOT touch the sliders! There is no check on max width / height for the input fields.
I think though that separate width and height input must be possible i.e. to turn off the aspect ratio restriction. I know - not the best for quality images but you may need images with the same size.
There is an issue for that. #483

schnuti:
And a last comment having nothing to do with this issue.
You can get the original size back after rotating using an additional step.
[...]

PR #496, opened 2018-01-16
Added distinct angle values for rotate

schnuti:
Pull Request for Issue #495 .

### Summary of Changes
- Added distinct angle values
- Removed the slider 
- Changed input to use a number field type
You can use the steps OR enter a value directly in a number field!

- Changed the quality fields to use number field type.

### Testing Instructions
Test that the angles are edited together with the image.

Test the quality fields in crop, resize and rotate view

The distinct values **are not changed ** when the number field is edited!

rotate


schnuti:
As in the cropping view the radio buttons do not get checked.

schnuti:
I messed things up a bit. I meant to keep the change of the quality fields to use the number field type apart from the distinct rotation values. But it didn't work. I hope you can test it together - if you want to use it.

PR #500, opened 2018-01-18
Adapt the crop ratio CSS selector

laoneo:
As the CSS selector got removed in #498. This pr adapts the styling.

### Before
image

### After
image


schnuti:
Also with this you can get funny results when the column get too narrow. I hope "floating" buttons can fix it.  
not-very-responsive

brianteeman:
Aspect ratio are usually written 4:3 not 4/3
I would change 2/3 to 3/2 so that all the ratios are in the same orientation
screenshotr10-56-32

screenshot from https://calculateaspectratio.com/

laoneo:
@brianteeman can you open a new issue for that, or even a new pr? Thanks.

brianteeman:
@laoneo ok

schnuti:
Do not yet merge this until cleared if the radio buttons remain. The old class might be needed for other buttons. I'm adding a PR for an alternative now.

laoneo:
Removed the button dependent code as it is not needed anymore because of the drop down from #503.

PR #501, opened 2018-01-18
Aspect Ratio values

brianteeman:
Aspect ratio are usually written 4:3 not 4/3
I would also change 2/3 to 3/2 so that all the ratios are in the same orientation
screenshotr10-56-32

screenshot from https://calculateaspectratio.com/

### DIFF
--- a/plugins/media-action/crop/form/crop.xml
+++ b/plugins/media-action/crop/form/crop.xml
@@ -64,9 +64,9 @@
 			class=""
 			default=""
 			>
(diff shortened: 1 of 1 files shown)


ciar4n (approved):

brianteeman:
Thanks

schnuti:
Is it meant to be 3:2 with the correct factor 1.5 or "portrait" with factor 0.666666666  for 2:3?
The other ones are landscape oriented.

Issue #502, opened 2018-01-19
Landscape and portrait aspect ratios


schnuti:
Isn't a choice between landscape and portrait aspect ratios needed for cropping?

Let's say you have a 16:9 image with a person in the middle. You may want to cut out an image 3:4 of this person. i.e. the aspect ratio is in "portrait" format.

As is, you have to do this "per hand" as 4:3 gives you a landscape format.

I've used "Base on width or height" as choice for my own cropping method but Landscape/Portrait are probably better understood.

(The aspect ratio 1:1 is missing. Often used for avatars, icons and more. I'll add it to another issue. )

laoneo:
I would leave them, perhaps we should switch to a drop down as mentioned here #501 where we can add a text.

schnuti:
What would you leave? What I try to say is, that you need double values for the aspect ratios. Yes, a drop down could replace the radio buttons.

Original aspect ratio (default as suggested by someone in another issue)
1:1
Landscape
5:4
4:3
3:2
16:9
Portrait
4:5
3:4
2:3
9:16
Free aspect ratio - no restrictions

This would be my complete list with the most common aspect ratios for images.

schnuti:
I've rebuilt the aspect ratios to a list field. I believe this is a better way to present the aspect ratios. I'll have a look if I'm able to modify the JavaScript as well. Or what do you think?
Image 1: The list field (width?), Image 2: The list, Landscape and Portrait are headers (disabled options)

![aspect-ratio_1](https://user-images.githubusercontent.com/956537/35166415-22fa0b08-fd52-11e7-8af4-aa70e1afa640.jpg)

![aspect-ratio_2](https://user-images.githubusercontent.com/956537/35166432-30f1d768-fd52-11e7-8cdb-5faf05ea4b4b.jpg)

schnuti:
See PR 503

Issue #499, opened 2018-01-17
Replace radio buttons with pseudo buttons


schnuti:
To keep the radio buttons on a to-do list after #494 and PR #498 I added this issue.

### Expected result
Radio buttons with the look of normal buttons.

laoneo:
I don't understand what exactly you want to change. Can you elaborate a bit more, perhaps with a screenshot?

schnuti:
We have to wait until core has merged a couple of pending PR's for radio buttons but then buttons can replace the radio buttons. I think the new MM had it in the early days.
I think there's a need for some special styling as I see problems with the width of the grouped buttons. They are not responsive.
Image shows an example from a frontend view I use in J3 showing 4 radio buttons as buttons.
 
![radio-as-button](https://user-images.githubusercontent.com/956537/35083570-1b5f882e-fc20-11e7-97f2-09192f476fe3.jpg)

laoneo:
Would be nice if you can link the related pr's here.

schnuti:
PRs and issues of interest in core (maybe more) 
joomla/joomla-cms#19326 - PR Fix radio btn group
joomla/joomla-cms#19322 - Rewrite markup radio form field
joomla/joomla-cms#19296 - PR Split the radio Field to 2 layouts

laoneo:
Thanks, so we wait?

schnuti:
Better do nothing regarding radio buttons until core has a basic solution or everything may have to be reverted/redone.

schnuti:
See PR 503

schnuti:
Let's keep this as a todo - reminder for the radio buttons on the rotate view.

schnuti:
Existing radio buttons resolved with #508

PR #503, opened 2018-01-20
Aspect ratio in crop view - Replace radio buttons with drop down list

schnuti:
Pull Request for Issue #502, #499 

@dgt41 
> For the other issue let’s add a first option: original aspect ratio which should be selected by default

### Summary of Changes
- Replaced the radio buttons with a drop down.
- Added a default to use original image aspect ratio
- Added Landscape and Portrait orientation alternatives
- Changed the Reset to read "No aspect ratio" for cropping without restrictions.

### Testing Instructions
Click on the different options in the drop down and use the Ratio aspects to crop the images.

Please check the Javascript. I'm not that fit and the Dom results has changed in 4.0

Image 1: The field, Image 2 Drop down opened. 
aspect-ratio_1
aspect-ratio_2


brianteeman:
No Aspect Ratio doesnt really make sense to me. Just "none" should do

Why do the two disabled options have a value?

media/plg_media-action_crop/js/crop.js
```diff
@@ -57,7 +57,16 @@ Joomla.MediaManager.Edit = Joomla.MediaManager.Edit || {};
 			Joomla.MediaManager.Edit.crop.cropper.setData({height: parseInt(this.value)});
 		});
 
-		var elements = document.querySelectorAll("#jform_aspectRatio input");
+		var elements = document.querySelectorAll(".crop-aspect-ratio-option");
```
dgrammatiko: Please make this an array: 
```js
var elements = [].slice.call(document.querySelectorAll(".crop-aspect-ratio-option"));
```

media/plg_media-action_crop/js/crop.js
```diff
@@ -57,7 +57,16 @@ Joomla.MediaManager.Edit = Joomla.MediaManager.Edit || {};
...
+
+		// Set default aspect ratio after numeric check, option has a dummy value
+		var defaultCropFactor = image.naturalWidth / image.naturalHeight;
+		if (!isNaN(+defaultCropFactor) && isFinite(defaultCropFactor)){
+			elements[0].value = defaultCropFactor;
+		}
+			Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(elements[0].value);
+
+		// Set click listener and function to the aspect ratio list
 		for (var i = 0; i < elements.length; i++) {
```
dgrammatiko: and change this for loop to a foreEach:
```js
elements.forEach(function(element) {
  element.addEventListener('click', function (e) {
   Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(this.value);
  });
});
```

dgrammatiko (changes requested):

schnuti:
@brianteeman 
Yes, None would probably be enough. I was only sure that Reset would not be adequate.

I had some minor problems with the list and think different dummy values are needed. i.e. only one "" is allowed, used for the "None" value.

schnuti:
@dgt41 Made an update with the changes. Thanks.

There seem to be a problem with the initialisation. The default is not always active after loading the image for editing. When you click on Reset in the toolbar it becomes active. Any idea? I'll have to debug that.

schnuti:
@dgt41 The problem shouldn't have with this PR to do.
The first load of an image sets wrong pos-x, pos-y, width and height. A reset sets them correct.

After first load
loaded_img

After reset
reset_img

I'll test some more to see if I can find the worm.

schnuti:
@dgt41 I found the problem. Pretty sure. An event is triggered as the Aspect ratio is changed to the new default.
`
Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(elements[0].value);
` 
My guess is that this line executes to early. But move to where?

media/plg_media-action_crop/js/crop.js
schnuti: done

media/plg_media-action_crop/js/crop.js
schnuti: done

schnuti:
I should mention that the values are show correct in Javascript debugging mode..

dgrammatiko:
probably it has to do with the event listener `change` (change is ok for select elements not for plain inputs), try setting a timeout of 200 ms in those listeners

schnuti:
@dgt41 
I was fishing in the wrong waters. The image takes it's time to load. 
`
var image = document.getElementById('image-preview');
`
How can this be handled? Some kind of "load"?
I set a timeout of 2000 but that can't be the solution. Can it be depending on filesize and web speed?

schnuti:
@dgt41 
I was fishing in the wrong waters. The image takes it's time to load.

`
var image = document.getElementById('image-preview');
`

I added a timeout of 2000 but that can't be the solution. Is the time depending on image size and web speed? Or would 200 be enough? Is there some kind of "load" check available.

schnuti:
@dgt41 
Now it works as expected.
Thanks!

dgrammatiko (approved):

laoneo:
Why is landscape and portrait disabled? If you want to make them grouping, probably you should use the group form field then.

schnuti:
@laoneo Sorry, wasn't aware of grouped list. I'll have to check some of my very old code where I probably should change to grouped list.
PR updated.

brianteeman:
Is anything supposed to happen when I select a ratio?
When I change a dimension then something happens but not when I select a ratio. See movie below

mm

schnuti:
Changing the aspect ratio should change the chosen area, though with "reset" values. A lot more logic would be needed to keep one of the values entered. Is the Javascript missing for you? Did it work before the drop down?
The movie I made became too large. I'll see if I can reduce the size.

schnuti:
The Movie. The first width input changes the height according to the "default" aspect ratio. i.e. the loaded images width and height ratio.

aspect_ratio

brianteeman:
It was a completely clean install from your branch

schnuti:
Where you have included this PR? 
If nothing happens when you chose another aspect ratio I suppose the crop.js file isn't working as by me.
Some other tester?

schnuti:
To check if something might be missing I made a clean install from my own patch-2 branch including this patch.
Everything works fine and as expected.

schnuti:
@brianteeman 
A closer look! You are probably using the first version of this PR. I can see, that you don't have the grouped list. Before that, a Javascript problem was discussed above and resolved. You'd better use the latest version.
(For a quick test with the old version, you can click once on the reset button in the toolbar. This should activate the aspect ratio options. )

brianteeman:
This PR is your branch - patch-2
That is what I tested

schnuti:
@brianteeman  Sorry for almost not believing you. I think I found out that some code is browser dependent. I'll have to give another developer tool a try, favourite is Firefox.

Not a defence - but it looks like IE11 (latest automatically updated version on my Win10) is not supported at all by MM.

@dgt41 Do you have an idea what could be browser dependent in the code?
image.addEventListener('load' ? The aspect ratio options do obviously not get initialised correct.

brianteeman:
I can confirm it works with firefox but not with chrome

dgrammatiko:
@schnuti I have no clue why chrome is failing here, try to debug `console.log(image.naturalWidth);` and similarly for the naturalHeight. Is it 0? try maybe `parseInt(image.naturalWidth)` 

About IE11, probably we're missing couple polyfills there (for sure the promise one)

schnuti:
@dgt41 
The code is going through correct but the events for the options are not "active" in Chrome.

`
    // Set click listener and function to the aspect ratio list

    elements.forEach(function(element) {
        element.addEventListener('click', function (e) {
             Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(this.value);
` 

The line addEventListener runs for all option elements in Chrome **and** Firefox.
The click event setAspectRatio only in FF.
My guess is that the option elements are tricky. The old replaced radio buttons are ok in both Chrome and FF. 
A bit desperate I tested the old code with no better result

`
for (var i = 0; i < elements.length; i++) {
`

schnuti:
@dgt41  It looks as I was right. Chrome and Safari ignores click events on select options.
What do you say about those changes? Tested with FF and Chrome.
1. Catch the change event on the parent (select)

```
    document.getElementById('jform_aspectRatio').addEventListener("change", function (e) {
        Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(this.value);
    });
```

2. Change the default option value

```
    // Wait for the image to load its data
    image.addEventListener('load', function() {
        
        // Get all option elements if future need
        var elements = [].slice.call(document.querySelectorAll(".crop-aspect-ratio-option"));
       
        // Set default aspect ratio after numeric check, option has a dummy value
        var defaultCropFactor = image.naturalWidth / image.naturalHeight;
           
        if (!isNaN(+defaultCropFactor) && isFinite(defaultCropFactor)){
            elements[0].value = defaultCropFactor;
        }
        Joomla.MediaManager.Edit.crop.cropper.setAspectRatio(elements[0].value);
    }); 
```

laoneo:
@schnuti can you commit your fix that we can merge it?

schnuti:
OK! Fix commited.

schnuti:
A tip if you like to use the keyboard. Use "tab" to go from the input fields to "Default aspect ratio". Use the up-, down-, left-, right-keys to change the aspect ratio. An effect of the last change.

PR #505, opened 2018-02-07
Bug mimetype in resize.js and rotate.js

schnuti:
Use the same code as in crop.js for the file format (used for the mime type)
myimage.jpeg becomes png encoded as jpg is invalid.
I couldn't find out if this has any consequences but it sure could.


laoneo:
Thanks!

Issue #507, opened 2018-02-08
Proposal: Add flip horizontal and flip vertical to the image rotate view.


schnuti:
With only a few lines the flip functionality can be added to the image crop view. This is often used in particular if the base is an analogue image negative or diapositive where up and down/left and right easily gets incorrect. 
Flip is included in the existing rotate events. A click on 0 degrees flips the image without rotation.
See gif animation. I'll add a PR later.

![flip-flop](https://user-images.githubusercontent.com/956537/35973452-ca450a1a-0cd5-11e8-80cb-1f8e6f92a916.gif)

laoneo:
Would be a nice addition!

schnuti:
PR added

laoneo:
We keep the issues open till it is finished and merged.

PR #511, opened 2018-02-08
Add flip functionality to the image crop view

schnuti:
Pull Request for Issue #507 

_Description from the issue_
With only a few lines the flip functionality can be added to the image crop view. This is often used in particular if the base is an analogue image negative or diapositive where up and down/left and right easily gets incorrect. 
Flip is included in the existing rotate events. A click on 0 degrees flips the image without rotation.
See gif animation. I'll add a PR later.

flip-flop


schnuti:
It would be nice to have buttons for the check boxes like the radio buttons.
There was a javascript for this in J3 but I think this has bin removed with no replacement for field type checkboxes.

dgrammatiko:
@schnuti there will be a custom element for that. This also looks good, great work. Thanks

dgrammatiko (approved):

laoneo:
I like the flip functionality, but for me it is confusing to have the check boxes connected to the degrees. Should we not just add two new buttons to flip the image horizontally and vertically?

schnuti:
I have a solution where events are added to the checkboxes. The function is the same but as you don't have to click on or change the angel it might be less confusing.
What do you say about the non existing correlation between entered angle and the "quick" buttons. As i understand, it's a bit tricky and the code would run for each and every click on the arrows in the number field. A "softer" way is to use the same css for active and not active buttons. The active state is only used when you click on a button anyway. The valid angel comes from the number field.

laoneo:
I would disconnect the flip actions from the angle. If you click on flip horizontally it flips it. That's all.

schnuti:
That's not all. The logic has to be changed. As is, each iteration , both rotate and flip, starts with the image loaded at init. To keep this, the flip functionality must then get it's own plugin and that's overkill.

laoneo:
I fear that how it is now, nobody will understand it. Just checked gimp and there you don't have to set a flag and then rotate it. Perhaps it makes code wise sense how it is now, but from user perspective it is not clear.

laoneo:
I mean codewise, nothing should stop you to add a new listener and do the canvas stuff there.

laoneo:
Where do we stay here?

laoneo:
@schnuti any plans to proceed with this pr?

laoneo:
@schnuti any plans to proceed with this pr?

kasvith:
@schnuti have plans to move with pr ?

schnuti:
As you do not accept it as is. No change awaiting. Use the idea if you like.

PR #528, opened 2018-04-09
Feature : Search Functionality in media manager

geetanshu-m:
### Summary of Changes

Added an Input field in the toolbar - media manager, which can filter the items based on the keyword provided. It would be searching the file/folder in the current directory only.


laoneo:
Thank you. Guess you need to clean up the files as the diff in the ApiController is a whole replacement of the file.

geetanshu-m:
Sorry For that, raised another PR #529

Issue #430, opened 2017-10-03
Copy public file url to clipboard


laoneo:
A feedback on Jday Germany was to copy the public file url of a media file to clipboard.

kasvith:
we need a toolbar button for that

dneukirchen:
add an icon-button next to the path in the media infobar.

kasvith:
This should be work only on files IMO

kasvith:
@dneukirchen i dont agree adding a button next to the path, since this should be explicitly work on files rather than folders, so my suggestion is to popup an icon on files or when a file is selected show a tool bar icon to copy url

kasvith:
another suggestion is to add a button before delete icon

laoneo:
I would also add it to the action buttons here
![image](https://user-images.githubusercontent.com/251072/34664874-4941a520-f45e-11e7-8ffc-2dccf76dd570.png)

kasvith:
yeah agreed, rather than creating a toolbar button this is better

dneukirchen:
i do not agree here. We should not put so many actions in the quick button menu. At least in its current form it is meant to hold only the most important actions.

ciar4n:
I with David on this, already feels like one to many actions.  I'm not against placing the action here, just think we need to reconsider how they are displayed.

kasvith:
hmm, then we have to enable this button on toolbar when user clicks only a single file. 
Let me know when you guys finalized this. Putting icon on file makes it simple, but as mentioned it overwhelm the file icon too.
 
let me know @dneukirchen @laoneo @ciar4n

ciar4n:
For the time been I'd say add it here...

![image](https://user-images.githubusercontent.com/2803503/34719227-d860a12e-f531-11e7-9619-4d1828033562.png)

I think we just need to reconsider how these icons are displayed which can be done after.

laoneo:
I would also not start spreading the actions all over the UI which are related to an item. Perhaps we need to reconsider the action drop down, but actions which do belong to an item (file, image, video, etc.) should appear close to it. At the moment it is the drop down.

dneukirchen:
I do not fully agree, but im ok with it. 

imo copy path to clipboard is different than the primary actions like rename, remove, delete, move... its a more like a convenience helper. Imo Inline menus should be 100% intuitive and users should not have to search for the action they want to use... We should also keep in mind that we need the inline menu in the table view as well.

laoneo:
[...]

PR #531, opened 2018-04-13
Feature - Copy url of a file to clipboard

kasvith:
Pull Request for Issue #430  .

### Summary of Changes
- Added a new button to the action list
- Added ability to copy the public url of a file to the clipboard with vanillaJS
- Added localizable strings for new feature (Default texts can be changed in language files)

### Testing Instructions
- Open NMM
- Open Action list of a file as follows and click on the link icon
image
- Click on **Get sharable link** button
image
- After some time you will get the sharable url of the file as follows
image
- You can copy it using the button next to text box or copy it your own

### Expected result
- Url of the file should be in your clipboard
- Try pasting it out

### Actual result
- It works 💯 

- [document.execCommand('copy')](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) is supported by IE9+ and others, The text box is there if browser does not support copy API
- I've not added any accessibility setups


dneukirchen:
Thank you kasun. It looks good, but for me it feels a little weird to get a sharable link for a static file on local filesystem. Imo this is another feature where the cloud plugins overcomplicate things. 

UX/UI is a matter of taste. Like posted in the original issue, i would prefer to have the copy/paste input in the infobar on the right and get rid of modals.

kasvith:
@dneukirchen yeah i would like that too because in the list view we will not have this feature.

basically this should be same for every adapter, make them more generalized, sharing a local file also would be nice if we adding ACL features to certain files. 

For cloud and other adapters this is preferred.

kasvith:
@laoneo what's your idea on that?

laoneo:
Please revert the change in the banner.jpg. If @dneukirchen has approved it, I will merge it.

kasvith:
@laoneo done, did not notice that lol

laoneo:
For a second part (doesn't need to be done in this pr), we can skip the get shareable link step for local files.

kasvith:
hmm...but i still prefer to have it as we don't get full url here by default. It's generated by the API

laoneo:
That's why it is better to do it in a second pr, to not block this one.

kasvith:
Yeah we have the basic functionality in here, then its just adjusting it

laoneo:
Conflicts again

kasvith:
Yeah i will rebuild scripts as Search box was merged

kasvith:
Resolved conflicts

administrator/language/en-GB/en-GB.com_media.ini
```diff
@@ -67,3 +68,6 @@ COM_MEDIA_TOGGLE_INFO="Toggle item info"
 COM_MEDIA_TOGGLE_LIST_VIEW="Toggle between grid and table views"
 COM_MEDIA_TOGGLE_SELECT_ITEM="Toggle select item"
 COM_MEDIA_UPDLOAD_SUCCESS="Item uploaded."
+COM_MEDIA_SHARE="Share the file"
```
laoneo: We have the policy in Joomla to alpha order the strings.

kasvith: Will fix asap

kasvith:
@laoneo done it

laoneo:
Very good job, I see we have a new vue.js master in our team!

kasvith:
Still learning from a master @dneukirchen

Issue #469, opened 2017-11-18
[4.0Alpha] Its TOO easy to delete whole folders


PhilETaylor:
When I clicked a folder, and then clicked delete, it deleted the whole folder, and all images, with no confirmation, and no warning.... gone... all my images.... (ok so this was a test only, but idiots will delete everything and blame Joomla, so this has to be made more difficult for a user to destroy their site, at least warn them if deleting a folder recursively !!!)

PhilETaylor:
https://github.com/joomla/joomla-cms/issues/20367

laoneo:
This is a duplicate of #437.

PR #533, opened 2018-04-15
Fix/delete item alert

kasvith:
Pull Request for Issue #469  .

### Summary of Changes
- Added prompts before deleting a file or folder or selected items
- Localization files are added

### Testing Instructions
- Try to delete some folder or a file, it will ask you before action

### Expected result
- Ask dialog to be shown

### Actual result
- It works, after you confirmed it will take action

- I dont like the ask dialog which browser provide, but that's a joomla core thing


kasvith:
Some problem with my setup. will open later

Issue #452, opened 2017-10-26
Select all


brianteeman:
I am assuming that the checkbox before the breadcrumbs is a select all.

I would expect it to be checked ONLY when everything is selected and not checked only when nothing is selected.

The current behaviour is the opposite
<img width="778" alt="screenshotr20-25-25" src="https://user-images.githubusercontent.com/1296369/32072834-f97cf0fa-ba8b-11e7-96b4-59c4fafba6b6.png">

ciar4n:
I'd agree with you on this. Considering this is a toggle is makes sense that the icon represents the current state.

brianteeman:
especially as that is the behaviour throughout the rest of the joomla ui

laoneo:
The first implementation was like that in #363. But after playing with it, we agreed to change it to the opposite in #373.

brianteeman:
Then you changed it wrongly. If you look at every other "select all" checkbox in the joomla ui the unselected state is an unchecked box. Also from an a11y perspective it must be unchecked for unselected otherwise you run into all sorts of issues trying to override the default behaviour of assistive technology

dneukirchen:
thx.

PR #536, opened 2018-04-21
Select all check box correction

geetanshu-m:
Pull Request for Issue #452 .

### Summary of Changes

Select all checkbox is now checked when all the items are selected.


kasvith:
Seems we've get rid of confusion :)

administrator/components/com_media/resources/scripts/components/toolbar/toolbar.vue
```diff
@@ -51,7 +51,7 @@
                 return (this.isGridView) ? 'fa fa-list' : 'fa fa-th';
             },
             toggleSelectAllBtnIcon() {
-                return (this.allItemsSelected) ? 'fa fa fa-square-o' : 'fa fa-check-square-o'
+                return (this.allItemsSelected) ? 'fa fa-check-square-o' : 'fa fa fa-square-o'
```
kasvith: why two `fa`

geetanshu-m: I just exchanged ```fa fa fa-square-o``` and ```fa fa-check-square-o``` 
do one ```fa``` need to be removed?

kasvith: @ciar4n

laoneo:
If I'm not wrong we were discussing that at the last code sprint and decided to go the way how it now.

ciar4n:
I think it has changed a couple of times. Personally I agree with this PR. Makes sense that the icon represents the current state.

kasvith:
Makes sense for me too

dneukirchen:
For me too. It was confusing before. Thx @proman24

kasvith:
thanks @proman24

Issue #537, opened 2018-04-21
Media content view not scrollable 


anuragteapot:
### Steps to reproduce the issue
1. Upload large amount of images to MM.
2. Try access all image or try to scroll.

### Expected result
Media content view scrollable. You can access all files easily .

### Actual result
Media content view not scrollable. Can't access all files.

![screenshot from 2018-04-21 14-45-39](https://user-images.githubusercontent.com/30978328/39087494-75affcee-45bf-11e8-9ef7-1ff7f82de7f6.png)

### System information (as much as possible)
Linux

anuragteapot:
I try to do PR for this.

kasvith:
What you mean by try accessing all files?

anuragteapot:
> What you mean by try accessing all files?


We cant access all files because  <code> div </code> is not scrollable

anuragteapot:
@kasvith we need a behavior like this if we have large amount of images . 

![peek 2018-04-22 00-48](https://user-images.githubusercontent.com/30978328/39087961-289dd036-45c7-11e8-9d96-d882685fd4a0.gif)

anuragteapot:
But we can't scroll .
@ciar4n

kasvith:
Got it, it need to be fixed

kasvith:
I've assigned you for this.

ciar4n:
PR created.. #541

PR #541, opened 2018-04-22
Fix scroll in browser

ciar4n:
Pull Request for Issue #537 .

### Summary of Changes
Fixes the scroll bar in image browser.


laoneo:
Can you fix conflicts here?

laoneo:
When the browser window has a scroll bar (eg. in debug mode), then I have with this patch two scroll bars. Can it be changed the way that only one scroll bar will appear?

ciar4n:
Just removed the min height which resolves this issue.

administrator/components/com_media/resources/styles/components/_layout.scss
```diff
@@ -34,7 +34,6 @@
 .media-main {
   position: relative;
   height: calc(100vh - 160px);
```
laoneo, line 36: It still does not work, but when I remove this line then it is ok.

ciar4n: Ok. I see now. The issue is only present in debug mode.

Removing this line means the grid/list area does not scroll. All images are displayed and instead you use the default browser scroll. Originally  the idea to adding the scroll to the grid/list area was so you can scroll the images while still keeping the toolbar and sidebar displayed. Should I still remove it?

laoneo: You are the CSS master. I do just report what I see.

laoneo:
@ciar4n can you have another look on this pr. Something screwed up here.

ciar4n:
The SCSS compiler is only compiling to mediamanager.css however the media manager is loading mediamanager.min.css ?

Not the problem here but an issue I face when I attempt to fix this.

ciar4n:
All should be good now. 

There is a scroll bar on both the media browser (presuming enough items) and the viewport when debug is enabled but there is no sensible way to avoid that.

laoneo:
I still have two scroll bars
image

ciar4n:
Have you debug enabled?

laoneo:
Yes. Does that mather?

ciar4n:
It makes a difference here...

> There is a scroll bar on both the media browser (presuming enough items) and the viewport when debug is enabled but there is no sensible way to avoid that.

I could work on some sort of a solution however to me it is a bit of an edge case?

laoneo:
Ah no, when it is only when debug is enabled, then all is good. Now we need to fix the drone tests too.

Issue #527, opened 2018-04-01
Its possible to rename files and folders to existing files/folders


dneukirchen:
### Steps to reproduce the issue

Rename a file to an existing file or a folder to an existing folder

### Expected result

Rename should not be possible

### Actual result

Rename is possible.

kasvith:
hmm...this involes move in local adapter, i will take a look, at last time unit test passed

kasvith:
As Unit tests passing need to consider more about this....pretty new to this system testing

dneukirchen:
The failing system tests are currently skipped. You don't have to worry about them. Once we fixed the issue, I can enable them.

PR #542, opened 2018-04-30
Fix file get renamed to an existing file

anuragteapot:
Pull Request for Issue #527

### Summary of Changes
Changes in <code>ApiController</code>
Initially, set <code>$force</code> false.
To not force to overwrite

### Testing Instructions

Rename a file to an existing file or a folder to an existing folder

### Expected result

Rename should not be possible

### Actual result
Rename is possible.

### DIFF
--- a/administrator/components/com_media/Controller/ApiController.php
+++ b/administrator/components/com_media/Controller/ApiController.php
@@ -281,11 +281,11 @@ class ApiController extends BaseController
 
 			if ($move)
 			{
-				$destinationPath = $this->getModel()->move($adapter, $path, $destinationPath, true);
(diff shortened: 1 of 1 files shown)


anuragteapot:
@dneukirchen Can you please review ?

kasvith:
Ok this seems fine. It can be merged.
For some weird reason i ve forced it there by default. 

We need some option to whether force or not for move and copy.
Even though rename uses **move** we need to distinguish it with the others as for any reason rename should not be forced. Others can be forced on request in my opinion.

kasvith (approved):

kasvith:
But i would wait until system tests are back online before merging it

laoneo:
Thanks!

PR #547, opened 2018-05-22
Cleanup build

laoneo:
Deletes the invalid files from the repo and adds the minified files as /media/com_media/js/mediamanager.min.js and /media/com_media/css/mediamanager.min.css.

`npm run build` in /administrator/components/com_media will now look like

image

Additionally the map files are ignored, because they should not be in a SCM.


yvesh (approved):

dgrammatiko:
Does it even make sense to have maps for js?
That file is huge and it's totally useless as you need to enable the dev mode to get the vue helper, so it's just useless IMHO

laoneo:
I would leave it, I mean they are not commited to git and do stay only local after a npm run build.

Issue #560, opened 2018-06-04
Colourful bug in notifications.js


schnuti:
### Steps to reproduce the issue
Not that easy. You have to find a code that sets Notifications and edit the options. As is, you cant change colours. At least not with the used CSS.

Its based on option.type and not option.level.
e,g. option.type ="danger" for error messages.

All is now using info colours.

dneukirchen:
thx

PR #563, opened 2018-06-08
Notifications type for level in Notifications.js

schnuti:
Pull Request for Issue #560 .

### Summary of Changes
Notifications type for level in Notifications.js

@dneukirchen Someone has to rebuild the vue things I suppose.
Then test with some kind of error message. Should be "danger"-colours, Is now the same as the "success" i.e. is defaulted to colour "info".


schnuti:
Got this totally wrong!

PR #570, opened 2018-07-16
Use proper grid markup while editing

laoneo:
### Summary of Changes
The edit options are shown below the image canvas while editing an image. but they should be on the side. This is broken since the new template.


laoneo:
The forms are not properly hidden if the markup get changed.

PR #572, opened 2018-07-16
Make it IE11 compatible!

dgrammatiko:
Pull Request for Issue # .

### Testing Instructions
Apply this patch and check if IE11 is working as expected


kasvith:
i got this error from chrome 
image

administrator/components/com_media/tmpl/media/default.php
```diff
@@ -23,7 +23,15 @@
 HTMLHelper::_('behavior.keepalive');
 
 // Add javascripts
-HTMLHelper::_('script', 'media/com_media/js/mediamanager.js');
+//JHtml::_('script', 'media/com_media/js/mediamanager.js');
+
+Factory::getDocument()->addScriptDeclaration('
+document.addEventListener(\'WebComponentsReady\', function() {
+	var elem = document.createElement(\'script\');
+	elem.src = "' . \Joomla\CMS\Uri\Uri::root(false) . 'media/com_media/js/mediamanager.js";
```
dgrammatiko, line 31: @kasvith add
```js
elem.setAttribute("type", "text/javascript");
```
to pass the csp

kasvith: wait

kasvith: sure

kasvith:
changed to **mediamanager.min.js** and it works in chrome, **not in IE**

IE 11:
image

dgrammatiko:
Maybe we need to update the web component polyfills

dgrammatiko:
@kasvith can you do one more test here, please?

So pull this branch and then go to 
`administrator/templates/atum/index.php`
and insert this
```
<script src="https://cdnjs.cloudflare.com/ajax/libs/core-js/2.5.7/core.min.js"></script>
```
before the `<jdoc:include type="head" />`

Does the media manager works correctly?
Does the editing also works?

kasvith:
@dgrammatiko still not working and i get this in IE
image

Here is template
image

administrator/components/com_media/tmpl/media/default.php
dneukirchen: This feels wrong and hacky. Why do we need to use WebComponentsReady?

dgrammatiko: Why? The polyfills are loaded after DOMContentLaoded (lazyloaded with checks on the client support) thus we cannot use the old-fashioned DOM ready event. To use the DOM ready it means that we need to check the client capabilities in the server (sniffing the UA in the request) which has some very obvious limitations and for sure will be more expensive performance wise

dneukirchen: Yes, I know. But this is because of the way we are loading/using the polifills. 
Other solutions: 

1) (you are going to hate me for this one) deliver all polyfills that are needed for joomla min browser support by default (and let the User decide how much he wants to improve performance/drop polyfills).

2) Deliver polyfills for media manager in media manager and do not expect joomla to handle this for us.

3) Rename the event name (which is imo kind of missleading) to something better. Like PolyfillsLoaded/Ready or even JoomlaLoaded/Ready, so that we decouple polyfills from webcomponents.

dgrammatiko: Well, No 1 is the worse solution but the easiest to adopt (let's not do that)

N2 we're gonna have duplicate code one way or another because we also need these polyfills for the web components

N3 Probably this should be named as the standard Joomla event `joomla:updated` so I can see a `joomla:ready` event. Probably I need to spend some time and make a method similar to Joomla.webComponents() that would handle the loading of the polyfills (eg extend the code we already have and refactor it so we could pass the polyfill and the required check in the json, or something similar)

My point of view here is that I don't want to make polyfills a hard dependency as at some point in the life cycle of Joomla 4 IE11 will be deprecated/unused and then all these will be just bloated code...

brianteeman: IE11 is supported by MS until the end of support for windows 10 which is 2025 :(

mbabker: #FML

dgrammatiko: Although what Brian wrote is correct I know that major companies have a target to drop IE 11 by 2020!

brianteeman: We can only go by the official dates ;)

dgrammatiko: @brianteeman was that the case for IE8, IE9 and IE10? I don't remember if people stop supporting those before or after EOL

brianteeman: yes we can only go by the official dates under current policy

mbabker: For now as a project we are best suited making decisions based on support of the official upstream project/service.  So we should plan for IE11 support lasting at a minimum as long as it is officially supported by Microsoft, for better or worse.  From a practical perspective, if we find that usage in 2020 or 2021 has dwindled low enough before its end of support that trying to support it makes little to no sense, then the discussion can be had.  But that also requires a change to the project's development strategy similar to https://github.com/joomla/joomla-cms/issues/21025 (which I'm not going to hold my breath over being accepted, we're already stuck to the basically dead PHP 7.0 minimum and that officially dies in 4 months).

dgrammatiko: > and that officially dies in 4 months

Lovely 😃

dgrammatiko:
@dneukirchen please can you take over this one?

